PR #2, opened 2024-01-24
Add Model 2 protonfix

rawdatafeel:
Add a protonfix file for the Model 2 emulator, https://segaretro.org/Model_2_Emulator

Can PR the database with relevant info if this is merged :)


GloriousEggroll:
can you change the filename to ulwgl-modal2?

We implemented parsing of the second part of the name so that if it determines it's numeric it is parsed as a steam ID, this is needed so that valve's proton can apply it's fixes in the proton launch script.

for games not on steam the second part of the ID should have at least one letter but preferably be a phrase thats easily understandable simply so that it's not parsed as a steam id and proton steam id specific fixes therefore dont get applied

GloriousEggroll:
> > can you change the filename to ulwgl-modal2?
> > We implemented parsing of the second part of the name so that if it determines it's numeric it is parsed as a steam ID, this is needed so that valve's proton can apply it's fixes in the proton launch script.
> > for games not on steam the second part of the ID should have at least one letter but preferably be a phrase thats easily understandable simply so that it's not parsed as a steam id and proton steam id specific fixes therefore dont get applied
> 
> Sorry about that, done!
> 
> Does this mean that I can use model2 as my ID when launching Model 2 using the gamelauncher.sh?

yes ulwgl-modal2 specifically. i updated the database guidelines with an explanation and example of what i was talking about with steam's game specific fixes in proton

PR #11, opened 2024-02-07
Create ulwgl-2016590.py

nmlynch94:
Adding gamefix for Dark and Darker.

### DIFF
--- a/gamefixes-ulwgl/ulwgl-2016590.py
+++ b/gamefixes-ulwgl/ulwgl-2016590.py
@@ -1,0 +1,19 @@
+""" Game fix for Dark and Darker
+"""
+#pylint: disable=C0103
+
+import os
+from protonfixes import util
+
+def main():
+    """ Installs wininet and urlmon in order to allow Blacksmith Launcher to properly install the game.
+        
+        This also has the side effect of breaking voip in DungeonCrawler.exe, so we add registry entries
+        to only use native wininet and urlmon for Blacksmith.exe, and not DungeonCrawler.exe.
+    """
+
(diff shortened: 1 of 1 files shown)


GloriousEggroll:
Dark and Darker has a steam version: 

https://steamcommunity.com/app/2016590

The file name needs to follow naming guidelines:

https://github.com/Open-Wine-Components/ULWGL-database?tab=readme-ov-file#rules-for-adding-ulwgl-id-entries

Note #2:

2. If a game is on steam, the ULWGL ID will be ulwgl-steamid. If it is not, you can make up any value as long as it does not already exist.

nmlynch94:
Thank you. Sorry about that, I did read those naming conventions, but Dark and Darker doesn't show on the store when you search it. 

I knew they weren't allowed on Steam due to an ongoing lawsuit, but it turns out they were there briefly and then removed, leaving their steam ID even though you can't search it. I've updated the PR.

GloriousEggroll:
> Thank you. Sorry about that, I did read those naming conventions, but Dark and Darker doesn't show on the store when you search it.
> 
> I knew they weren't allowed on Steam due to an ongoing lawsuit, but it turns out they were there briefly and then removed, leaving their steam ID even though you can't actually get it from the store. I've updated the PR.

thanks! looks good now. can you also make an entry here:
https://github.com/Open-Wine-Components/ULWGL-database

nmlynch94:
> 
> thanks! looks good now. can you also make an entry here: https://github.com/Open-Wine-Components/ULWGL-database

Done! https://github.com/Open-Wine-Components/ULWGL-database/pull/11

PR #14, opened 2024-02-17
Add ZP platform and Duke MP fix

doZennn:
https://github.com/Open-Wine-Components/ULWGL-database/pull/17


GloriousEggroll:
as noted in the database PR the ulwgl-id must be tied to the steam id, we use 1 id per game, regardless of version and/or store differences. the differences are handled per-script not per ID. we use the store value to differentiate between which script should be used for the game's ID

PR #15, opened 2024-02-17
[Fix] GOG's Gothic 1 and 2

imLinguin:
fixes gothic 1 and 2 on gog. I don't have the 3'rd one so I'm unable to try that at the moment.

I also added new protonfixes utilities for getting file name case-insensitively.
I also made `get_game_install_path` actually return install path.

### DIFF
--- a/util.py
+++ b/util.py
@@ -351,10 +351,12 @@ def del_environment(envvar):
 def get_game_install_path():
     """ Game installation path
     """
-
-    log.debug('Detected path to game: ' + os.environ['PWD'])
+    install_path = os.environ['PWD']
+    if 'STEAM_COMPAT_INSTALL_PATH' in os.environ:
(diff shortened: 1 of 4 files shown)


util.py
```diff
@@ -557,7 +599,7 @@ def set_dxvk_option(opt, val, cfile='/tmp/protonfixes_dxvk.conf'):
     conf = configparser.ConfigParser()
     conf.optionxform = str
     section = conf.default_section
-    dxvk_conf = os.path.join(get_game_install_path(), 'dxvk.conf')
+    dxvk_conf = os.path.join(os.environ['PWD'], 'dxvk.conf')
```
imLinguin, line 602: I was not sure about this behavior so I replaced this with PWD env var like it used to be before get_game_install_path changes

GloriousEggroll:
lgtm, I have all 3 on steam, will double check and make adjustments if anything is funky and let you know

PR #16, opened 2024-02-20
gog: add Trails in the Sky 1-3

keenanweaver:
Adds the cutscene fix for The Legend of Heroes: Trails in the Sky 1-3, and preps for the [SoraVoice mod](https://www.pcgamingwiki.com/wiki/The_Legend_of_Heroes:_Trails_in_the_Sky#Add_Voice_Acting). If 'non-essential' changes aren't desired, I can remove that from the commit.

What's the consensus on these sorts of changes where the game can work without the change, but it's probably going to be done by the player anyway? (Assuming it doesn't inadvertently affect the game for people that don't install the mod.)

### DIFF
--- a/gamefixes-gog/ulwgl-251150.py
+++ b/gamefixes-gog/ulwgl-251150.py
@@ -1,0 +1,1 @@
+../gamefixes-steam/251150.py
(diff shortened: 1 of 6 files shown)


GloriousEggroll:
lgtm, ty!

PR #19, opened 2024-02-21
Add Street Racing Syndicate

doZennn:
https://github.com/Open-Wine-Components/ULWGL-database/pull/21

### DIFF
--- a/gamefixes-steam/292410.py
+++ b/gamefixes-steam/292410.py
@@ -1,0 +1,9 @@
+""" Street Racing Syndicate
+"""
+#pylint: disable=C0103
+
+from protonfixes import util
+
+def main():
+    util.protontricks('lavfilters') # fix videos
+    util.winedll_override('d3d9', 'n,b') # in case user uses the ThirteenAG widescreen fix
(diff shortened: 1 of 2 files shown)


R1kaB3rN:
Hi, thanks for your contribution

But I'm curious -- what exactly is wrong with the in-game videos that requires `lavfilters`?

gamefixes-steam/292410.py
```diff
@@ -0,0 +1,9 @@
+""" Street Racing Syndicate
+"""
+#pylint: disable=C0103
+
+from protonfixes import util
+
+def main():
+    util.protontricks('lavfilters') # fix videos
```
R1kaB3rN, line 8: I don't own this game to test, but if the video problems can be solved via `util.disable_protonaudioconverter()` that would be preferred as it would require no additional codecs to be installed into the prefix

doZennn: Videos look like this without `lavfilters` and `util.disable_protonaudioconverter()` doesn't do anything.
Screenshot_from_2024-02-22_12-19-38

PR #30, opened 2024-03-04
Add Github Action for protonfixes

R1kaB3rN:
Use Pylint to lint fixes and check for valid symbolic links


R1kaB3rN:
cc @imLinguin for GH Action permission -- it's not enabled

imLinguin:
Enabled now, might need another push to trigger, or having the PR as ready

PR #44, opened 2024-03-09
fix linting error

marianoag:
syswow64 = os.path.join(util.protonprefix(), 'drive_c/windows/syswow64', 'dgvoodoo.conf')

in https://github.com/Open-Wine-Components/ULWGL-protonfixes/pull/43 for some protonfixes the name of the file was added to syswow64 and it's fine but in this case syswow64 dir is used with another file so:

shutil.copy(os.path.join(syswow64, 'dgd3d9.dll'),os.path.join(syswow64, 'd3d9.dll'))
is drive_c/windows/syswow64/dgvoodoo.conf/d3d9.dll instead of drive_c/windows/syswow64/d3d9.dll 

@R1kaB3rN is this done automatically when the protonfix is linted or was this manual?

If it was manual it was a mistake there's no problem, but if it was automatic there's something wrong and maybe there are more protonfixes affected.

i guess it was manual because it wasn't modified now.

### DIFF
--- a/gamefixes-steam/409090.py
+++ b/gamefixes-steam/409090.py
@@ -12,7 +12,7 @@ import shutil
 from protonfixes import util
 
 def main():
-    syswow64 = os.path.join(util.protonprefix(), 'drive_c/windows/syswow64', 'dgvoodoo.conf')
+    syswow64 = os.path.join(util.protonprefix(), 'drive_c/windows/syswow64')
     if util.protontricks('dgvoodoo2'):
-        subprocess.call([f"sed -i '/[DirectX]/ {{/Resolution/s/max/unforced/}}' {syswow64}"], shell=True)
+        subprocess.call([f"sed -i '/[DirectX]/ {{/Resolution/s/max/unforced/}}' {syswow64}/dgvoodoo.conf"], shell=True)
     shutil.copy(os.path.join(syswow64, 'dgd3d9.dll'),os.path.join(syswow64, 'd3d9.dll'))


R1kaB3rN:
> syswow64 = os.path.join(util.protonprefix(), 'drive_c/windows/syswow64', 'dgvoodoo.conf')
> 
> in #43 for some protonfixes the name of the file was added to syswow64 and it's fine but in this case syswow64 dir is used with another file so:
> 
> shutil.copy(os.path.join(syswow64, 'dgd3d9.dll'),os.path.join(syswow64, 'd3d9.dll')) is drive_c/windows/syswow64/dgvoodoo.conf/d3d9.dll instead of drive_c/windows/syswow64/d3d9.dll
>
> @R1kaB3rN is this done automatically when the protonfix is linted or was this manual?
> 
> If it was manual it was a mistake there's no problem, but if it was automatic there's something wrong and maybe there are more protonfixes affected.
> 
> i guess it was manual because it wasn't modified now.

Thanks for reporting this. Yes, rest assured, this was manually edited and the linter will **not** automatically apply fixes. The reason for this particular change was to satisfy Pylint's f-string rule (C0209), which the first argument for the subprocess statements had violated.

This was actually one of the few cases that were, comparatively, non-trivial as for the rest I was mainly deleting trailing new lines or deleting/moving module statements. Other fixes that are similar to this include gamefixes-steam/302370.py which does not include a copy statement.

marianoag:
> Other fixes that are similar to this include gamefixes-steam/302370.py which does not include a copy statement.

And another one is 497360 (syswow64 also not used somewhere else) but you put /dgvoodoo.conf in the f-string instead of syswow64. the declaration was some lines above. When i cleaned the mess left after trial and error with protonfixes like these, somehow syswow was left outside the if, far from the sed call.

This was good for me, i learnt something new, didn't know about f-string.

PR #46, opened 2024-03-17
Added fix for Command and Conquer - Red Alert 2

Root-Core:
This fix has multiple purposes:

1. It fixes the black screen rendering issue.
2. It allows to install CnCnet Launcher and patches.
3. It runs the CnCnet Launcher, if installed.

I wasn't sure if it was okay to just install the Launcher without user consent. Therefore I implemented a toggle to request the installation. We could change this to default, the Launcher is de-facto the way to play the game these days.

I updated [cnc-ddraw in Winetricks](https://github.com/Winetricks/winetricks/pull/2199), but it isn't merged yet.
Feel free to cherry-pick.

### DIFF
--- a/gamefixes-steam/2229850.py
+++ b/gamefixes-steam/2229850.py
@@ -1,0 +1,41 @@
+""" Game fix for Command & Conquer Red Alert™ 2 and Yuri’s Revenge™
+"""
+#pylint: disable=C0103
+
+import os
+
+from protonfixes import util
+from protonfixes.logger import log
+
+def main():
+    """ Launch CnCNet Launcher - if installed, install it if requested
+
+        The game crashes, when the launcher and it's patches are installed
+        and Ra2.exe or RA2MD.exe are called directly by Steam.
+        For this reason we must always use the launcher, if present.
+
+        The game will just show a black screen without cnc-ddraw or the patch in place.
+    """
+
+    # User requested to install the CnCnet Launcher
+    if os.environ.get('INSTALL_CNCNET'):
+        log('User wants to install CnCnet Launcher')
+        util.protontricks('cncnet_ra2')
+
+    # CnCnet Launcher is in place, run it
+    if os.path.isfile('CnCNetYRLauncher.exe'):
+        log('CnCnet Launcher found, bypass game execution!')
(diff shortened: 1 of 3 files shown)


gamefixes-steam/2229850.py
```diff
@@ -0,0 +1,41 @@
...
+        log('User wants to install CnCnet Launcher')
+        util.protontricks('cncnet_ra2')
+
+    # CnCnet Launcher is in place, run it
+    if os.path.isfile('CnCNetYRLauncher.exe'):
+        log('CnCnet Launcher found, bypass game execution!')
+        util.replace_command('Ra2.exe', 'CnCNetYRLauncher.exe')
+        util.replace_command('RA2MD.exe', 'CnCNetYRLauncher.exe')
+    else:
+        # Return early, if cnc_ddraw is installed or failed to install
```
R1kaB3rN, line 31: I'm assuming you've tested this using Proton 9? Also, are there are any other benefits that the CnCnet launcher provides for users who choose not to use it?

Root-Core: > I'm assuming you've tested this using Proton 9?

I've tested it against Proton 8, but I could test it against 9. I don't think this should make a huge difference.

> Also, are there are any other benefits that the CnCnet launcher provides for users who choose not to use it?

It patches a lot of issues / crashes or works around them.
For example it allows windowed mode and other resolutions / aspect ratios than 640x480, 800x600, 1024x768.
The scroll speed on the screen edge is sometimes unplayable fast, it is configurable in the launcher.

Non-steam versions can run without a CD, which might also be in the scope of this project.

R1kaB3rN: Let's just make this opt-in until the [upstream tracking issue is closed](https://github.com/ValveSoftware/Proton/issues/6208) or if there's ever an official update that resolves the problems. This game looks to be practically borked out of the box and I wouldn't want users to have to do the tinkering themselves.

Root-Core: > Let's just make the CnCnet launcher the default
Done.

> This game looks to be practically borked out of the box
Unfortunately. EA could have used a working ddraw implementation, but they did not.
Even on Windows it doesn't exactly work perfectly.

PR #48, opened 2024-03-24
Refactoring: fix.py / linting code files / added Types

Root-Core:
Refactor:
- Typed most functions
- Major refactoring of fix.py
- Some minor refactoring

Linter:
- Added root to files to lint
- Fixed C0114, by adding docstrings to modules
- Removed linter exception for C0114
- Added check for timeout of urllib

### DIFF
--- a/checks.py
+++ b/checks.py
@@ -1,9 +1,6 @@
 """ Run some tests and generate warnings for proton configuration issues
 """
 
-import shutil
-import os
-import subprocess
 from .logger import log
 
 
--- a/checks.py
+++ b/checks.py
@@ -16,7 +16,7 @@ def esync_file_limits() -> bool:
     https://github.com/zfigura/wine/blob/esync/README.esync
     '''
 
-    with open('/proc/sys/fs/file-max') as fsmax:
+    with open('/proc/sys/fs/file-max', encoding='ascii') as fsmax:
         max_files = fsmax.readline()
         if int(max_files) < 8192:
             log.warn(warning)
--- a/checks.py
+++ b/checks.py
@@ -7,7 +7,7 @@ import subprocess
 from .logger import log
 
 
-def esync_file_limits():
+def esync_file_limits() -> bool:
     """
     Check esync file limits using /proc/sys/fs/file-max
     https://www.reddit.com/r/SteamPlay/comments/9kqisk/tip_for_those_using_proton_no_esync1/
@@ -26,7 +26,8 @@ def esync_file_limits():
             return False
     return True
 
-def run_checks():
+
+def run_checks() -> None:
     """ Run checks to notify of any potential issues
     """
 
--- a/download.py
+++ b/download.py
@@ -26,17 +26,19 @@ def gdrive_download(gdrive_id: str, path: str) -> None:
     cjar = http.cookiejar.CookieJar()
     opener = urllib.request.build_opener(urllib.request.HTTPCookieProcessor(cjar))
     urllib.request.install_opener(opener)
+
     req = urllib.request.Request(url)
-    resp = urllib.request.urlopen(req)
-    confirm_cookie = [x for x in resp.getheaders() if
-                      (x[0] == 'Set-Cookie'
-                       and x[1].startswith('download_warning'))][0][1]
+    with urllib.request.urlopen(req, timeout=10) as resp:
+        confirm_cookie = [x for x in resp.getheaders() if
+                        (x[0] == 'Set-Cookie'
+                        and x[1].startswith('download_warning'))][0][1]
     confirm = confirm_cookie.split(';')[0].split('=')[1]
-    req2 = urllib.request.Request(url + '&confirm={}'.format(confirm))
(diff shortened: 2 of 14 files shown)


R1kaB3rN:
Thanks, this is great, and I'm open to merging this. Though, do you mind breaking this down into smaller commits? It'll help with debugging/bisecting in case there's a potential bug from the refactoring.

Root-Core:
Yeah, sorry about that. I intended to do smaller commits, but somehow I ended up doing too much at once and it was hard to break it up into separate commits afterwards. Most changes are trivial, as types are not enforced by Python.

I will try to split it up, that might take some time.

checks.py
```diff
@@ -1,13 +1,10 @@
...
 """
 
-import shutil
-import os
-import subprocess
 from .logger import log
 
 
-def esync_file_limits():
+def esync_file_limits() -> bool:
```
R1kaB3rN, line 7: Hmm if we're adding types to non-gamefixes, type annotations should probably be enforced by the linter, Pylint, and caught in the CI system to avoid having to add the missing type in the future. This can be addressed in a separate PR though

See https://pypi.org/project/flake8-annotations/

Root-Core: We could also enforce types for the gamefixes.
This should be trivial, as most of them only consist of the `main()` function.

download.py
```diff
@@ -11,35 +11,37 @@
...
+def get_filename(headers: list) -> str:
     """ Retrieve a filename from a request headers via Content-Disposition
     """
     content_disp = [x for x in headers if x[0] == 'Content-Disposition'][0][1]
     raw_filename = [x for x in content_disp.split(';') if x.startswith('filename=')][0]
     return raw_filename.replace('filename=', '').replace('"', '')
 
 
-def gdrive_download(gdrive_id, path):
+def gdrive_download(gdrive_id: str, path: str) -> None:
```
R1kaB3rN, line 22: This functionality was inherited from upstream and IIRC it's not being used anywhere in this project. 

Personally, I haven't verified if it actually works either, and if possible, I'd like to remove this functionality all together. But since users might be using it in their local fixes, I guess we can't really do that.

fix.py
```diff
@@ -30,168 +33,158 @@ def game_id():
...
+
+
+def _run_fix_local(game_id: str, default: bool = False) -> bool:
+    """ Check if a local gamefix is available first and run it
     """
+    localpath = os.path.expanduser('~/.config/protonfixes/localfixes')
+    module_name =  game_id if not default else 'default'
+
+    # Check if local gamefix exists
+    if not os.path.isfile(os.path.join(localpath, module_name + '.py')):
```
R1kaB3rN, line 137: `f'{module_name}.py'`

download.py
Root-Core: I have come across a number of functions that I am convinced are not being used. I left them in and updated them anyway, as it wasn't in the scope of this PR.

Some stuff could be broken up, like certain functions of the `utils.py`.. but then again some users may need to update their local fixes.

fix.py
Root-Core: There are more of these, I updated strings that are formatted or consist of multiple concatenations.
I left strings that are simple prefixes or suffixes.

I could change them too, and we could also enforce them through the linter.

download.py
```diff
@@ -11,35 +11,37 @@
...
+
+    req = urllib.request.Request(f'{url}&confirm={confirm}')
+    with urllib.request.urlopen(req, timeout=10) as resp:
+        filename = get_filename(resp.getheaders())
+        with open(os.path.join(path, filename), 'wb') as save_file:
+            save_file.write(resp.read())
 
 
-def sha1sum(filename):
+def sha1sum(filename: str) -> str:
```
R1kaB3rN, line 44: Same with this function. It's sorta useless and it makes more sense if the winetricks executable uses it. But unfortunately it's a shell script, and for the same above reasons, we'll probably just have to leave it

checks.py
R1kaB3rN: We could and it'll be keep things consistent. Though, I don't think it'll add much though so you don't need to. 

For the game fixes, what we really want to catch from the linter are the typos, logical errors and some bad practices. There were just a ton of them before I added the linter in this project, and for the subset of fixes that return a non-None type for some reason I don't think it'll affect the behavior of the program

Root-Core: Sure, I think it is just easier to maintain one linter configuration than multiple. Alternatively there might be an option to allow functions that return `None` to have no type. Not sure about that though.

download.py
R1kaB3rN: For this specifically, if it can be proved that the google drive functionality does not work, then that would warrant removing it. Of course, it doesn't have to be addressed in this PR.

In a future separate PR, we can definitely evaluate removing some of them too. Or like, wait until the upcoming new year for the cleaning 😄

Root-Core:
@R1kaB3rN I split the commit into several, I hope they seem logical to you.

PR #49, opened 2024-03-28
Refactored replace_command()

Root-Core:
Sometimes updates change the capitalization of game executables.
This is not a problem on windows, but breaks fixes on Linux.

- Allow case insensitive matching
- Returns if there was a match
- Logging respects if there was a match

### DIFF
--- a/util.py
+++ b/util.py
@@ -335,14 +335,28 @@ def regedit_add(folder,name=None,type=None,value=None,arch=None):
     process = subprocess.Popen(regedit_cmd, env=env)
     process.wait()
 
-def replace_command(orig_str, repl_str):
-    """ Make a commandline replacement in sys.argv
-    """
 
-    log.info('Changing ' + orig_str + ' to ' + repl_str)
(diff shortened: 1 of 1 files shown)


util.py
```diff
@@ -342,14 +342,26 @@ def regedit_add(folder: str, name: str = None, typ: str = None, value: str = Non
...
+    found = False
     for idx, arg in enumerate(sys.argv):
-        if orig in arg:
-            sys.argv[idx] = arg.replace(orig, repl)
+        replaced = re.sub(orig, repl, arg, flags=match_flags)
+        if replaced == arg:
+            continue
+        sys.argv[idx] = replaced
+        found = True
+
```
R1kaB3rN, line 359: The loop can be terminated after a match was found, unless I'm missing something? Otherwise the loop will probably continue processing the game/launcher's arguments if there's any.

Root-Core: I'm not sure tbh. I just replicated the previous behaviour, which does the full loop.

PR #54, opened 2024-04-04
Cease to Breathe fix

marianoag:
Replace included nwjs(0.71) wich doesn't work with 0.86
Fix cursor hitbox (set frame=false in package.json)
Updated from 0.85 that didn't display custom cursors.

### DIFF
--- a/gamefixes-steam/1873170.py
+++ b/gamefixes-steam/1873170.py
@@ -1,0 +1,28 @@
+""" Cease to Breathe
+Replace included nwjs(0.71) wich doesn't work with 0.86
+Fix cursor hitbox (set frame=false in package.json)
+Updated from 0.85 that didn't display custom cursors.
+"""
+#pylint: disable=C0103
+
+import os
+import glob
+import shutil
+import urllib.request
+import zipfile
+import subprocess
+from protonfixes import util
+
+def main():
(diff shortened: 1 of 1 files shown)


gamefixes-steam/1873170.py
```diff
@@ -0,0 +1,28 @@
...
+import urllib.request
+import zipfile
+import subprocess
+from protonfixes import util
+
+def main():
+    util.replace_command('CTB.exe', 'nw.exe')
+    install_dir = glob.escape(util.get_game_install_path())
+    if not os.path.isfile(os.path.join(install_dir, 'nw.exe')):
+        url = 'https://dl.nwjs.io/v0.86.0/nwjs-v0.86.0-win-x64.zip'
```
R1kaB3rN, line 20: At this endpoint, the digest file [SHASUMS256.txt](https://dl.nwjs.io/v0.86.0/SHASUMS256.txt) is included. Please verify that the digest matches for this executable somewhere then I'll merge this ASAP.

And in the case it doesn't match, I would just log a warning and not apply the fix.

marianoag: Checks zip file sha256 hash and logs error to console if it fails.

PR #55, opened 2024-04-04
Ceville

marianoag:
stopped working even with same proton version it used to work.
pipeline library error while loading.


gamefixes-steam/23460.py
```diff
@@ -15,3 +15,4 @@ def main():
     #util.protontricks('klite')
     if os.path.isdir('./data/shared/videos'):
         subprocess.call(['mv', './data/shared/videos', './data/shared/_videos'])
+    util.set_environment('WINEDLLOVERRIDES', 'libvkd3d-1=n')
```
R1kaB3rN, line 18: Prefer `util.winedll_override('libvkd3d-1', 'n')`

marianoag: Done.

PR #57, opened 2024-04-13
Game fix for Fallout 76

dunconio:
Ensure Faudio is installed so UI and NPC audio and music is audible.


R1kaB3rN:
https://github.com/ValveSoftware/Proton/issues/3766

PR #58, opened 2024-04-15
Sherlock Holmes Chapter One ultrawide

marianoag:
I made this to be able to use one of two UW fixes automatically. it supports:
[RoseTheFlower/SherlockHolmesCOUltrawide](https://github.com/RoseTheFlower/SherlockHolmesCOUltrawide)
[PhantomGamers/shco-wsf](https://github.com/PhantomGamers/shco-wsf)

First it checks that the display is UW (in case there was a UW fix but the display changed or the game files were copied for example to a SteamDeck) Then it checks if Rose fix is already present, after it checks for Phantom's and if neither is present it downloads Rose's one. The first is more complicated (it's a trainer) the other is an ASI loader.

For the trainer it checks for the game pid (made a function for that), to be sure it's running before launching the trainer in another thread, then it keeps checking every 5 seconds until the game is not running anymore so it can kill the trainer (otherwise Steam thinks the game is still playing)

Also it does another thing that is not necessary, but more convenient especially if playing with a gamepad, it minimizes the launcher window. For this it needs xdotool and libxdo.so3.

This is cumbersome it needs to download two .deb files and since /usr/bin/ar is not present it downloads busybox_AR to decompress them, then it decompresses the tar.xz inside (luckily with python this time) It would be great if xdotool and libxdo.so3 could be included alongside xrandr.

I hope it can be included even if it's not a fix, since it does nothing unless there's an UW monitor. I prefer not to but if needed i could remove both the atomatic download and the minimization of the trainer to remove the need for xdotool.

### DIFF
--- a/gamefixes-steam/1137300.py
+++ b/gamefixes-steam/1137300.py
@@ -17,7 +17,6 @@ import urllib.request
 import hashlib
 from time import sleep
 from protonfixes import util
-from protonfixes.logger import log
 
 def _get_pid(procname: str) -> int:
     procpid = None
@@ -57,7 +56,7 @@ def install_xdotool(path_to_exe: str):
         file_sum = hashlib.sha256(f.read()).hexdigest()
     if hashsum == file_sum:
         os.chdir(path_to_exe)
-        subprocess.Popen(['./busybox_AR', 'x', file, 'data.tar.xz'])
+        subprocess.call(['./busybox_AR', 'x', file, 'data.tar.xz'])
         os.chdir('../../..')
         with tarfile.open(f'{path_to_exe}/data.tar.xz', 'r') as zip_ref:
             for member in zip_ref.getmembers():
@@ -75,7 +74,7 @@ def install_xdotool(path_to_exe: str):
         file_sum = hashlib.sha256(f.read()).hexdigest()
     if hashsum == file_sum:
         os.chdir(path_to_exe)
-        subprocess.Popen(['./busybox_AR', 'x', file, 'data.tar.xz'])
+        subprocess.call(['./busybox_AR', 'x', file, 'data.tar.xz'])
         os.chdir('../../..')
         with tarfile.open(f'{path_to_exe}/data.tar.xz', 'r') as zip_ref:
             for member in zip_ref.getmembers():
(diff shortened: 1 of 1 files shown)


R1kaB3rN:
I appreciate the effort here to improve the game. But since this game seems to already be at a playable and this module is not fixing anything in the wine prefix that would bring it to a playable level for users, I believe it would be out of scope for this project...

Also, you're applying the mods by default for all users. So even if it's merged, I believe it would be best that you make users aware of it and allow them to enable each mod via an environment variable.

marianoag:
> I appreciate the effort here to improve the game. But since this game seems to already be playable and this module is not fixing anything in the wine prefix that would bring it to a playable level for users, I believe it would be out of scope for this project...

It's not exactly a fix, but it fixes something that cannot be easily done in Linux but it can be done in Windows. You cannot alt-tab and then doubleclick the trainer executable, you have to launch it with Proton in the same prefix. So you would need to make a script to run the trainer.
And in the case of the ASI loader it needs a dll override.

> Also, you're applying the mods by default for all users. So even if it's merged, I believe it would be best that you make users aware of it and allow them to enable each mod via an environment variable.

It doesn't apply the fix by default it checks for a resolution of 21:9 or wider first. Also the download of the fix could be eliminated and then only if the user downloaded the zip to the install folder it will be applied.

R1kaB3rN:
This isn't a fix and doesn't bring the game to a Playable level, so I don't plan to merge it in this project. I'll leave this open though, unless someone on the team thinks otherwise

This is more suited for another project that I plan on trying to get in GE in the future, since some game require mods or patches to either improve the gaming experience or had to be purposely left out due to Steam content restrictions

marianoag:
> This isn't a fix and doesn't bring the game to a Playable level, so I don't plan to merge it in this project. I'll leave this open though, unless someone on the team thinks otherwise

Like i said i know it doen't make a game playable, but there are 2 very different UW fixes:

Rose (trainer) What it does is launch the trainer, minimize it and close it when the game is finished. Compared to windows where you have to double click to launch the trainer, with proton to be able to do the same you would have to create a script where you launch the game using the game's prefix. if the goal is to behave the same way as Windows.
Minimize and close the trainer after the game ended is something that the Windows users would have to do just like us, this part is just an enhancement.

Phantom (loader) Unlike the other which i understand if it's not included, this behaves very differently. If the user copies the files to the binaries folder then it would just work in Windows. In Proton you need to override dsound.dll to make it work. And it's just 1 line (3 if you include a check for 21:9 or wider and if dsound.dll exists)

> This is more suited for another project that I plan on trying to get in GE in the future, since some games require mods or patches to either improve the gaming experience or had to be purposely left out due to Steam's content restrictions

This is interesting, i would like to know more about it. I suppose that it includes Gamescope too? (saw your drafts was waiting for it)

Also it would be great to have a compatibility tool that you can choose for any game, something that would check the game and the use your future project or Roberta or Boxtron or Luxtorpeda and if none targets the game then just use GEProton.

GloriousEggroll:
To put it bluntly if it doesn't allow the game to be playable I'd rather not merge such changes. If the game is already playable and this improves upon the functionality such as window behavior and gamepad input then I have no problem with it.

marianoag:
> To put it bluntly if it doesn't allow the game to be playable I'd rather not merge such changes. If the game is already playable and this improves upon the functionality such as window behavior and gamepad input then I have no problem with it.

tl;dr Even one line is better than nothing but i'll explain what they do and what can be removed.

This works for any of 2 different UW fixes:

1. The simple one is just one line, a dll override needed for the UW fix to use the dll.
2. The complex one allows to launch the game without having to create a script to have the UW fix applied (in Windows it's just double clicking the executable but because of Proton in Linux there's the need for a script to launch on the same prefix)

Why the big one if the other is just one line? Because it's better and also it just works if the user chose to use that one.
But it could be made smaller either by adding xdotool in GEProton (like xrandr) or not minimizing the trainer window.

Also what's nice to have but only has to be done once and can be removed is:
- Unzip fix if zip file is found.
- Download fix if UW resolution is detected.

Also this will never affect other users because it only applies if both UW resolution is detected and the files are present, so even if the user has the files in the prefix but is using another monitor or the SteamDeck then nothing will happen.

GloriousEggroll:
Someone else in our discord brought up introducing ultrawide/widescreen dll overrides -- I think a safe middleground for now would be gating them behind an envvar

PR #60, opened 2024-04-15
Many Video Playback fixes for Yuzusoft VN's

CreeperofDoom:
Allows WMP9 Video Playback to work for many Yuzusoft Steam titles

### DIFF
--- a/gamefixes-steam/1144400.py
+++ b/gamefixes-steam/1144400.py
@@ -1,0 +1,15 @@
+""" Game fix for Senren Banka
+"""
+#pylint: disable=C0103
+
+from protonfixes import util
+
+def main():
+    """
+    """
+    util.protontricks('quartz')
+    util.protontricks('wmp9')
+    util.protontricks('wmp9_x86_64')
+    util.protontricks('klite')
+    util.protontricks('directshow')
+    util.disable_protonaudioconverter()
--- a/gamefixes-steam/1277930.py
+++ b/gamefixes-steam/1277930.py
@@ -1,0 +1,15 @@
+""" Game fix for Riddle Joker
+"""
+#pylint: disable=C0103
+
+from protonfixes import util
+
+def main():
+    """
+    """
+    util.protontricks('quartz')
+    util.protontricks('wmp9')
+    util.protontricks('wmp9_x86_64')
+    util.protontricks('klite')
+    util.protontricks('directshow')
+    util.disable_protonaudioconverter()
--- a/gamefixes-steam/1829980.py
+++ b/gamefixes-steam/1829980.py
@@ -1,0 +1,15 @@
+""" Game fix for Cafe Stella
+"""
+#pylint: disable=C0103
+
+from protonfixes import util
+
+def main():
+    """
+    """
+    util.protontricks('quartz')
+    util.protontricks('wmp9')
+    util.protontricks('wmp9_x86_64')
+    util.protontricks('klite')
+    util.protontricks('directshow')
+    util.disable_protonaudioconverter()
--- a/gamefixes-steam/2458530.py
+++ b/gamefixes-steam/2458530.py
@@ -1,0 +1,15 @@
+""" Game fix for Sanoba Witch FHD Edition
+"""
+#pylint: disable=C0103
+
+from protonfixes import util
+
+def main():
+    """
+    """
+    util.protontricks('quartz')
+    util.protontricks('wmp9')
+    util.protontricks('wmp9_x86_64')
+    util.protontricks('klite')
+    util.protontricks('directshow')
+    util.disable_protonaudioconverter()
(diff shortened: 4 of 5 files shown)


marianoag:
You should try to check what's really necessary:

both wmp9_x86_64 and wmp9? I don't even think that both wmp9 and klite are needed, in ProtonDB i saw wmp9 or wmp11, just choose one.

Then if quartz is not enough (maybe amstream also) but both of those are called by directshow in winetricks, so if you use directshow they are redundant.

CreeperofDoom:
> You should try to check what's really necessary:
> 
> both wmp9_x86_64 and wmp9? I don't even think that both wmp9 and klite are needed, in ProtonDB i saw wmp9 or wmp11, just choose one.
> 
> Then if quartz is not enough (maybe amstream also) but both of those are called by directshow in winetricks, so if you use directshow they are redundant.

I wanted to ensure it would work, i tried the necessary commands to make sure it would work. Quartz fixes the audio problem with the video playback and maybe the Proton Audio Converter script helped.

marianoag:
> I wanted to ensure it would work, i tried the necessary commands to make sure it would work. Quartz fixes the audio problem with the video playback and maybe the Proton Audio Converter script helped.

Sure it will work but it is overkill. To make it clean you should delete the prefix and try again, since it's impossible to uninstall a winetricks verb.

If i were you i would try quartz and wmp9, if it doesn't work try directshow and if it works remove quartz.

If not delete the prefix again and try quartz and klite, and the directshow as above if it didn't work.

CreeperofDoom:
> > I wanted to ensure it would work, i tried the necessary commands to make sure it would work. Quartz fixes the audio problem with the video playback and maybe the Proton Audio Converter script helped.
> 
> Sure it will work but it is overkill. To make it clean you should delete the prefix and try again, since it's impossible to uninstall a winetricks verb.
> 
> If i were you i would try quartz and wmp9, if it doesn't work try directshow and if it works remove quartz.
> 
> If not delete the prefix again and try quartz and klite, and the directshow as above if it didn't work.

Trying Directshow, Quartz and Klite result in the same nothing as before. Trying Quartz and Klite result in a plain black screen, trying wmp9 itself brings an error, trying wmp9_x86_64 brings and error. If i try all the above (which is the one i tried to publish) it will work normal

marianoag:
> Trying Directshow, Quartz and Klite result in the same nothing as before. Trying Quartz and Klite result in a plain black screen, trying wmp9 itself brings an error, trying wmp9_x86_64 brings and error. If i try all the above (which is the one i tried to publish) it will work normal. changing the order does nothing. I made sure to clear the Prefix before trying each new one. and trying quartz, wmp9 and wmp9_x86_x64 results in video but broken audio

That's odd that you need two wmp9. Also odd that i didn't see that in ProtonDB but also  it's  common to see reports saying that everything works fine but the cutscenes are not displayed or there's no audio.

If you haven't tried or you don't remember i bet that just directshow + wmp9 and possibly disable protonaudioconverter for the missing audio could work.

CreeperofDoom:
> > > I wanted to ensure it would work, i tried the necessary commands to make sure it would work. Quartz fixes the audio problem with the video playback and maybe the Proton Audio Converter script helped.
> > 
> > 
> > Sure it will work but it is overkill. To make it clean you should delete the prefix and try again, since it's impossible to uninstall a winetricks verb.
> > If i were you i would try quartz and wmp9, if it doesn't work try directshow and if it works remove quartz.
> > If not delete the prefix again and try quartz and klite, and the directshow as above if it didn't work.
> 
> Trying Directshow, Quartz and Klite result in the same nothing as before. Trying Quartz and Klite result in a plain black screen, trying wmp9 itself brings an error, trying wmp9_x86_64 brings and error. If i try all the above (which is the one i tried to publish) it will work normal. changing the order does nothing. I made sure to clear the Prefix before trying each new one. and trying quartz, wmp9 and wmp9_x86_x64 results in video but broken audio

Alright i got it. Quartz, WMP9 and Directshow + DisableProtonAudioconverter (if its not there the audio doesn’t play) If i try it with Directshow and without WMP9 it doesn’t work, if i try it without quartz and with directshow it doesn’t work. Please, trying all this out was stressful enough and i just want 

> > Trying Directshow, Quartz and Klite result in the same nothing as before. Trying Quartz and Klite result in a plain black screen, trying wmp9 itself brings an error, trying wmp9_x86_64 brings and error. If i try all the above (which is the one i tried to publish) it will work normal. changing the order does nothing. I made sure to clear the Prefix before trying each new one. and trying quartz, wmp9 and wmp9_x86_x64 results in video but broken audio
> 
> That's odd that you need two wmp9. Also odd that i didn't see that in ProtonDB but also it's common to see reports saying that everything works fine but the cutscenes are not displayed or there's no audio.
> 
> If you haven't tried or you don't remember i bet that just directshow + wmp9 and possibly disable protonaudioconverter for the missing audio could work.

Alright i got it. Quartz, WMP9 and Directshow + DisableProtonAudioconverter (if its not there the audio doesn’t play) If i try it with Directshow and without WMP9 it doesn’t work, if i try it without quartz and with directshow it doesn’t work. Please, trying all this out was stressful enough

PR #62, opened 2024-04-16
(New) Many WMP9 Video Playback fixes for Yuzusoft VN's

CreeperofDoom:
This will fix 100% WMP9 Video Playback for many Yuzusoft VN's, Including Riddle Joker. I am very certain this works, excluding even a single command will either cause a black screen, laggy or no audio or it just wont play. (I tested this very well)

### DIFF
--- a/gamefixes-steam/1144400.py
+++ b/gamefixes-steam/1144400.py
@@ -1,0 +1,13 @@
+""" Game fix for Senren Banka
+"""
+#pylint: disable=C0103
+
+from protonfixes import util
+
+def main():
+    """
+    """
+    util.protontricks('quartz')
+    util.protontricks('wmp9')
+    util.protontricks('directshow')
+    util.disable_protonaudioconverter()
--- a/gamefixes-steam/1277930.py
+++ b/gamefixes-steam/1277930.py
@@ -1,0 +1,13 @@
+""" Game fix for Riddle Joker
+"""
+#pylint: disable=C0103
+
+from protonfixes import util
+
+def main():
+    """
+    """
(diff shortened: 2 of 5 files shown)


CreeperofDoom:
Failed? Nothing is wrong with it, what am i supposed to change here?

R1kaB3rN:
Thanks for the pull request. I'll test them in a bit since I own them, and I think it would probably be a good idea to create fixes in GOG or umu because they're available in other storefronts -- GOG and NekoNyan

gamefixes-steam/1144400.py
```diff
@@ -0,0 +1,13 @@
+""" Game fix for Senren Banka
+"""
+#pylint: disable=C0103
+
+from protonfixes import util
+
+def main():
+    """
+    """
```
R1kaB3rN, line 9: It's the empty docstring which is unnecessary. See the linter errors for more details

R1kaB3rN:
To clarify, this fixes the video playback specifically for the opening and ending videos, correct?

CreeperofDoom:
> To clarify, this fixes the video playback specifically for the opening and ending videos, correct?

Yes, ive tested. In the normal playthrough they play with full video and sound. Replaying them through the scene select also has full video and sound

gamefixes-steam/1144400.py
CreeperofDoom: and does this mean i have to submit another pull request with those gone in the script

CreeperofDoom:
i will now submit yet another Pull Request with the spaces gone so it can pass the checks.

PR #63, opened 2024-04-16
(Newer) Many WMP9 Video Playback fixes for Yuzusoft VN's

CreeperofDoom:
This will fix 100% WMP9 Video Playback for many Yuzusoft VN's, Including Riddle Joker. I am very certain this works, excluding even a single command will either cause a black screen, laggy or no audio or it just wont play. (I tested this very well) Reuploaded again to fix a tiny error, Only the Steam editions of these titles were tested and Full audio and video for FMV Scenes work. Other storefronts will most likely work considering there is no difference between versions

### DIFF
--- a/gamefixes-steam/1144400.py
+++ b/gamefixes-steam/1144400.py
@@ -1,0 +1,13 @@
+""" Game fix for Senren Banka
+"""
+#pylint: disable=C0103
+
+from protonfixes import util
+
+def main():
+    """
+    """
+    util.protontricks('quartz')
+    util.protontricks('wmp9')
+    util.protontricks('directshow')
+    util.disable_protonaudioconverter()
--- a/gamefixes-steam/1277930.py
+++ b/gamefixes-steam/1277930.py
@@ -1,0 +1,13 @@
+""" Game fix for Riddle Joker
+"""
+#pylint: disable=C0103
+
+from protonfixes import util
+
+def main():
+    """
+    """
+    util.protontricks('quartz')
+    util.protontricks('wmp9')
+    util.protontricks('directshow')
+    util.disable_protonaudioconverter()
--- a/gamefixes-steam/1829980.py
+++ b/gamefixes-steam/1829980.py
@@ -1,0 +1,13 @@
+""" Game fix for Cafe Stella
+"""
+#pylint: disable=C0103
+
+from protonfixes import util
+
+def main():
+    """
+    """
+    util.protontricks('quartz')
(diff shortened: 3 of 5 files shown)


marianoag:
Great job. That's much better now, not worth to keep trying. Although it's a shame that i don't have any of those games because i'm curious where the problem is with quartz since it's called by directshow. Maybe calling directshow before wmp9 would work, if not then it's a problem with the order that amstream is called inside directshow.

By the way did you figure out the naming of the protonfix for the non steam game and the entry for the database?

CreeperofDoom:
> Great job. That's much better now, not worth to keep trying. Although it's a shame that i don't have any of those games because i'm curious where the problem is with quartz since it's called by directshow. Maybe calling directshow before wmp9 would work, if not then it's a problem with the order that amstream is called inside directshow.
> 
> By the way did you figure out the naming of the protonfix for the non steam game and the entry for the database?

I figured the Non-Steam one out, The title i tried was Sabbat of the Witch directly from the Nekonyan Store (Doing it was a hard process requiring Launch commands and a directory different from the normal GE-Proton one) I am not going to bother with retrying the scripts i submitted because if it already works then there isn't a reason to keep trying.

marianoag:
> > I am not going to bother with retrying the scripts i submitted because if it already works then there isn't a reason to keep trying.

Yes, there's no need. I was just curious. It's not like quartz is installed twice, when directshow calls it it's already installed.

R1kaB3rN:
Presumably, you must be creating the files from Github instead from a terminal, correct? Because you didn't have to close the previous pull request.

Also, have you tested with `wmp11`?

gamefixes-steam/1144400.py
```diff
@@ -0,0 +1,13 @@
+""" Game fix for Senren Banka
+"""
+#pylint: disable=C0103
+
+from protonfixes import util
+
+def main():
+    """
+    """
```
R1kaB3rN, line 9: Either delete this empty docstring or add a short comment summarizing the fix and its aim. Otherwise, the CI will always fail.

gamefixes-steam/1144400.py
```diff
@@ -0,0 +1,13 @@
...
+    util.protontricks('quartz')
+    util.protontricks('wmp9')
```
R1kaB3rN, line 11: Have you tested with `wmp11`?

CreeperofDoom: WMP11 install instead of WMP9 results in a black screen.

PR #66, opened 2024-04-26
Fixes for Oddworld Abe's Oddysee & Stranger's Wrath HD, Overlord II, Ducati World Championship

doZennn:
I can split these up into multiple PR's if that's preferable.


R1kaB3rN:
> I can split these up into multiple PR's if that's preferable.

Yes, that would be preferred!

PR #71, opened 2024-04-28
Ducati World Championship

doZennn:


gamefixes-steam/6270.py
```diff
@@ -0,0 +1,8 @@
+""" Ducati World Championship
+"""
+#pylint: disable=C0103
+
+from protonfixes import util
+
+def main():
+    util.protontricks('icodecs')
```
R1kaB3rN, line 8: Can we possibly use another codec? This codec is known to be insecure by Microsoft and navigating to the url for that codec will display a warning when using a modern browser.

R1kaB3rN: Downloading the file from curl will also fail:

> curl: (60) SSL: no alternative certificate subject name matches target host name 'www.cucusoft.com'
> More details here: https://curl.se/docs/sslcerts.html
> 
> curl failed to verify the legitimacy of the server and therefore could not
> establish a secure connection to it. To learn more about this situation and
> how to fix it, please visit the web page mentioned above.

doZennn: No, tried a few other things before icodecs but there's no other way around using it unfortunately.

PR #74, opened 2024-05-07
Add fix for Total War: Shogun 2 on Steam

crumbfive:


crumbfive:
> How was the experience for you when using D3D9 instead of D3D11 by setting `PROTON_NO_D3D11=1`? Also, [according to a recent latest report](https://www.protondb.com/app/34330?device=any), the native application apparently runs perfectly when using D3D9.

I didn't try 'PROTON_NO_D3D11=1' but you can choose "Directx 9" when starting the game or from within the game which seemed to work ok both with and without the fix from my quick test. Using D3D11(with this fix) gives more graphical options in game and seems to look better. But again only a quick test/comparison.

R1kaB3rN:
https://github.com/ValveSoftware/Proton/issues/7110

PR #75, opened 2024-05-08
Add fixes for COJ Gunslinger on Steam

Tiagoquix:
- Add wmp11 to fix cutscenes
- Limit to 31 cores

Please let me know if I missed something. Thanks!

wmp11 and 31 cores fixes are due to user reports:

https://github.com/ValveSoftware/Proton/issues/366

https://www.protondb.com/app/204450

It seems like both `wmp9` and `wmp11` work, so we should prefer the newer version.
I am using `wmp11` and it works normally.

### DIFF
--- a/gamefixes-steam/204450.py
+++ b/gamefixes-steam/204450.py
@@ -1,0 +1,13 @@
+""" Game fix Call of Juarez: Gunslinger
+"""
+#pylint: disable=C0103
+
+from protonfixes import util
+
+def main():
+    """ installs wmp11
+    """
+    # Fixes missing audio cutscenes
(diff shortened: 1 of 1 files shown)


R1kaB3rN:
> @R1kaB3rN Hi there. Could you please clarify for me if the following is needed for future contributions? Thanks in advance.
> 
> ```python
> #pylint: disable=C0103
> 
> from protonfixes import util
> ```

The first statement is not strictly necessary when it comes to applying fixes, so users shouldn't need to worry about it. 

As for the second statement, it depends but you will most likely need it to do anything useful since all the functionality to apply a fix are currently in that file.

PR #77, opened 2024-05-11
star citizen gamefixes

mactan-sc:
remove un-needed overrides
add environment variable radv_zero_vram


GloriousEggroll:
radv_zero_vram should never be a permanent setting. if it needs to fix a graphics isssue then it needs to be report to upstream mesa

mactan-sc:
thanks for the feedback, holding off for upstream

PR #78, opened 2024-06-02
Add fix for Persona 4 Golden on Steam

OlegAckbar:


R1kaB3rN:
Do you mind providing more context for reinstating this fix? This was removed since 5188e666ac4f4132838e6163d38081e8176944a4, so it should be fine now.

@tgurr for feedback

OlegAckbar:
I guess the Steam version doesn't need it since it probably ships re-encoded videos, but if you try to run the game outside Steam environment (e.g. to make DRM-free version with Goldberg emu) you'll have to re-encode in-game videos on your own PC and that will faild without wmp9, klite etc. Also, the game relies on `d3dx11_43` which won't be installed since the Steam first-run process would never happen.

tgurr:
> @tgurr for feedback

I can only comment on the Steam version/usage and also not sure if this project is intended to cover niche use-cases such as the one mentioned above forcing unneeded stuff for the majority of people and/or if it could be made optional by detecting if running "under" steam or not. For running under/with Steam workarounds are certainly not needed anymore.

I also can't spot a dedicated separate depot for the re-encoded video files on https://steamdb.info/app/1113000/depots/ so shouldn't everyone get them by default when downloading the latest version? (The game went through some major updates over the time even years after the initial release with it even moving to 64bit, additional subtitles and so on.)

doZennn:
> I also can't spot a dedicated separate depot for the re-encoded video files on https://steamdb.info/app/1113000/depots/

Steam distributes transcoded media via the shader cache updates.

tgurr:
> shader cache update

Interesting, good to know thanks! Off-topic: Is there some special-case handling implemented? I can remember having disabled downloading shader cache globally but Persona 4 still worked for me including playback of video.

PR #79, opened 2024-06-04
Create 4730.py

wantija:
outrun 2006 coast 2 coast water fix


GloriousEggroll:
this is overwriting the d3dx9 dll which forces it to use native opengl. This should not be done, instead it should be reported to dxvk:

https://github.com/doitsujin/dxvk/

GloriousEggroll:
nvm, after a double check d3dx9 is for the entire group of d3dx9_* helper dlls, not d3d9.dll. this is ok. ty!

PR #81, opened 2024-06-10
Adding protonfix for Hogwarts Legacy

jwdb:
vcrun2019 is required for the game to start. Otherwise the game will throw an error on start.


imLinguin:
Hey, thanks for this. I noticed the fix is redundant for steam, so could you move the file from steam to epic directory, so it will be applied only for the epic version?

PR #83, opened 2024-06-21
Contractors VR

ToRRent1812:


R1kaB3rN:
I see that there are reports of this needing Microsoft Visual C++ Redistributable. However, currently, applying the `vcrun2022` verb shouldn't work as the digest had changed for some reason. See https://github.com/Winetricks/winetricks/issues/2235

The file of a digest being changed is serious and this needs to be fixed upstream before being merged.

GloriousEggroll:
since we ship winetricks we can change the digest in our shipped version for the time being

PR #84, opened 2024-06-26
gog: add fix for original resident evil

imLinguin:
cutscenes still don't work, but that may be a wine issue

### DIFF
--- a/gamefixes-gog/umu-1580232252.py
+++ b/gamefixes-gog/umu-1580232252.py
@@ -6,3 +6,4 @@ from protonfixes import util
 
 def main():
     util.winedll_override('ddraw', 'n,b')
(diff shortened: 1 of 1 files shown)


gamefixes-gog/umu-1580232252.py
```diff
@@ -0,0 +1,8 @@
+""" Resident Evil (1997)
+"""
+#pylint: disable=C0103
+
+from protonfixes import util
+
+def main():
+    util.winedll_override('ddraw', 'n,b')
```
keenanweaver, line 8:
```suggestion
    util.winedll_override('ddraw', 'n,b')
    util.winedll_override('dinput', 'n,b')
```
For the controller fixes.

PR #85, opened 2024-06-30
WRC 4 added

ToRRent1812:
Let's make it run out of the box.

### DIFF
--- a/gamefixes-steam/256330.py
+++ b/gamefixes-steam/256330.py
@@ -8,8 +8,12 @@ def main():
     #Black screen without it
     util.protontricks('d3dx9_42')
     util.protontricks('d3dx9_43')
-    util.protontricks('faudio')
+
+    #Fixes background videos
+    util.protontricks('wmp11')
+
+    #Fixes audio sliders in options
     util.protontricks('xact')
 
-    # Audio breaks above 60 fps
+    # Audio breaks above 60 fps, game engine issue
     util.set_environment('DXVK_FRAME_RATE', '60')
--- a/gamefixes-steam/256330.py
+++ b/gamefixes-steam/256330.py
@@ -14,6 +14,3 @@ def main():
 
     #Fixes audio sliders in options
     util.protontricks('xact')
-
-    # Audio breaks above 60 fps, game engine issue
-    util.set_environment('DXVK_FRAME_RATE', '60')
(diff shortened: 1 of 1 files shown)


gamefixes-steam/256330.py
```diff
@@ -0,0 +1,15 @@
...
+
+def main():
+    #Black screen without it
+    util.protontricks('d3dx9_42')
+    util.protontricks('d3dx9_43')
+    util.protontricks('faudio')
+    util.protontricks('xact')
+
+    # Audio breaks above 60 fps
+    util.set_environment('DXVK_FRAME_RATE', '60')
```
R1kaB3rN, line 15: Instead of this fix, a PR should be made for this upstream in DXVK.

gamefixes-steam/256330.py
```diff
@@ -0,0 +1,15 @@
...
+"""
+#pylint: disable=C0103
+
+from protonfixes import util
+
+def main():
+    #Black screen without it
+    util.protontricks('d3dx9_42')
+    util.protontricks('d3dx9_43')
+    util.protontricks('faudio')
```
R1kaB3rN, line 11: Why does this solve? Using the faudio verb will install downgraded faudio DLLs and use them. Are you sure this is necessary?

gamefixes-steam/256330.py
ToRRent1812: audio bug is not linux related, should I still ask DXVK to limit framerate upstream?

gamefixes-steam/256330.py
ToRRent1812: Tried without it, looks like it works now, without xact, audio sliders do not work in game. I also forgot to add wmp11 to the script to make background videos work.

gamefixes-steam/256330.py
R1kaB3rN: Yes, please make a PR to DXVK as everyone will benefit and you're certain that playing the game above 60+ FPS will break audio. Also, I'm not even too sure we allow fixes to set this variable as this fix would be only one that sets this environment variable in this project.

R1kaB3rN:
I'll merge this ASAP once you make a DXVK PR and remove that environment variable.

See https://github.com/doitsujin/dxvk/blob/master/src/util/config/config.cpp for examples.

PR #92, opened 2024-07-09
Fix: The Last Blade

ranplayer:


R1kaB3rN:
I read the upstream report and this looks to be graphics related. 

Is there any reason why you haven't reported this upstream in DXVK? Because having this fixed in DXVK would be more preferable than a fix in this project.

R1kaB3rN:
Sorry, never mind. I was looking into this issue and saw https://github.com/doitsujin/dxvk/issues/4095, so it appears to be a problem in wine.

gamefixes-steam/465840.py
```diff
@@ -0,0 +1,10 @@
+""" 
+The Last Blade
+"""
+#pylint: disable=C0103
+
+from protonfixes import util
+
+def main():
+    util.protontricks('d3dcompiler_46')
+
```
R1kaB3rN, line 10: After this trailing new line is deleted, I'll merge this ASAP.

ranplayer: done

Issue #89, opened 2024-07-07
Unify UPlay fixes


Tiagoquix:
https://github.com/Open-Wine-Components/umu-protonfixes/blob/783c82f1d3870fd623b5091abbe26206cb347946/util.py#L460

The disable_uplay_overlay fix is not shared for some fixes. Instead, they manually have the fix, and do not use this global function.

https://github.com/Open-Wine-Components/umu-protonfixes/blob/783c82f1d3870fd623b5091abbe26206cb347946/gamefixes-steam/812140.py#L16

https://github.com/Open-Wine-Components/umu-protonfixes/blob/783c82f1d3870fd623b5091abbe26206cb347946/gamefixes-steam/447040.py#L18

These should be changed to use the global function.

------------------------------------

The global function to disable overlay could also include the CloseBehavior change.
Here is what it needs:
```
user:
  closebehavior: CloseBehavior_Close
```
(taken from my Watch Dogs 1 install, as I had to manually change it)

See the Lutris install script for WD1, it changes the close behavior: https://lutris.net/games/install/25509/view (and also some of the fixes from protonfixes also change it)

PR #94, opened 2024-07-14
Unify UPlay fixes

R1kaB3rN:
Closes https://github.com/Open-Wine-Components/umu-protonfixes/issues/89.

cc @Tiagoquix for testing

### DIFF
--- a/gamefixes-steam/447040.py
+++ b/gamefixes-steam/447040.py
@@ -1,21 +1,11 @@
-""" Game fix for Watch_Dogs 2
-"""
-#pylint: disable=C0103
+"""Game fix for Watch_Dogs 2"""
+# pylint: disable=C0103
 
-import os
 from protonfixes import util
 
+
 def main():
-    """ disable Easy Anti-Cheat and online play, disable uplay overlay and change closebehavior
-    """
-
-    uplayconfigpath = os.path.join(util.protonprefix(), 'drive_c/users/steamuser/Local Settings/Application Data/Ubisoft Game Launcher')
-    if not os.path.exists(uplayconfigpath):
-        os.makedirs(uplayconfigpath)
-    uplayconfigfile = os.path.join(uplayconfigpath, 'settings.yml')
-    if not os.path.isfile(uplayconfigfile):
-        with open(uplayconfigfile,"w+", encoding='utf-8') as f:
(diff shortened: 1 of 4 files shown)


R1kaB3rN:
If you have a relevant game where this fix would applicable, you can use [umu-launcher](https://github.com/Open-Wine-Components/umu-launcher) and set the `GAMEID` to the fix's file name. Or alternatively, in the Steam client, set the `SteamAppId`.

R1kaB3rN:
cc @Tiagoquix for testing. Assuming it's always a good thing to do, the Uplay overlay will be disabled by default when possible.

R1kaB3rN:
> Hello, how can I test this specifically? Do I have to compile Proton or just downloading the `.py` files is fine?

It's simple. Just temporarily move the `protonfixes` directory with all of the files in this branch. Assuming you have GE-Proton9-11 installed:

1. Go to https://github.com/R1kaB3rN/ULWGL-protonfixes/tree/uplay-overlay
2. `git clone https://github.com/R1kaB3rN/ULWGL-protonfixes.git`
3. `mv $HOME/.local/share/Steam/compatibilitytools.d/GE-Proton9-11/protonfixes $HOME/.local/share/Steam/compatibilitytools.d/GE-Proton9-11/.protonfixes`
4. `mv ULWGL-protonfixes $HOME/.local/share/Steam/compatibilitytools.d/GE-Proton9-11/protonfixes`

Afterwards, just be sure to restore the directory.

R1kaB3rN:
> I can test on https://steamdb.info/app/470220/, but I would have to either create a script for it or set Steam App ID. How can I set Steam App ID?

You don't need to set one if you own that game. If you use my branch, the Uplay overlay for that game should be disabled.

R1kaB3rN:
So in this PR, I'm deciding not to disable uplay overlay by default due to the risk of core in-game functionality being lost for games. The goal of this project is bring games to a more playable state, not potentially cause any more issues (e.g., loss of multiplayer). This is something that should be handled case-by-case as only the developers of the game or its users would know whether the loss of functionality as a result would cause or not cause any issues.

Therefore, in cases where the uplay overlay would cause the game to be borked or severely detriment the gaming experience, I believe the appropriate action to take would be to create a fix for the game **and** report the problem upstream.

PR #96, opened 2024-07-16
Add GOG symlink for Beyond Good and Evil

imLinguin:
* Remove unnecessary dsound download
* ~~Set a DXVK framerate limit to 60~~


Blisto91:
Game already have a built in 60fps in dxvk. Is it not applying correctly?

imLinguin:
It is in fact applied correctly

PR #105, opened 2024-08-26
Improved and combined fixes for Gothic 1 / 2

Root-Core:
The current fixes are not sufficient to make the games playable.

- The mouse is not usable in both games without overriding `dinput` / `dinput8`
- Gothic II crashes on startup without `directmusic` installed
- Gothic I does not play background music without `directmusic` installed
- There are reports of audio issues without setting the latency
- This fix also allows the use of the most popular workshop mods
- Having `fsync` enabled doesn't cause any problems

As both games use the very same fix, I just symlinked them. I hope this is okay.


R1kaB3rN:
- https://github.com/ValveSoftware/Proton/issues/3245
- https://github.com/ValveSoftware/Proton/issues/1256

PR #107, opened 2024-08-27
xliveless.verb: Fix install location of xlive.dll

ProjectSynchro:
Previously this wouldn't install xlive.dll in any system folders, essentially making this fix do nothing.

I bumped the year, as well as the version of the ASI Loader as there's been quite a few releases since version 4.68.

Tested this with DiRT2 (A GFWL title) But I could also test on Fable 3 if that is something that would be needed. :)

### DIFF
--- a/verbs/xliveless.verb
+++ b/verbs/xliveless.verb
@@ -1,15 +1,15 @@
 w_metadata xliveless dlls \
     title="Games for Windows Live (gfw / gfwl) - xlive.dll mock, single player only" \
     publisher="ThirteenAG" \
-    year="2022" \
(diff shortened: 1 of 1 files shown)


ProjectSynchro:
Yes, the existing verb doesn't seem to work with DiRT2. I think this is down to the GFWL redistributable being installed as a setup step.

I tested this verb with Fable 3 (I don't own Lost Planet 2 unfortunately) other games that included the verb should work fine as even if they install GFWL, xliveless is installed over it.

R1kaB3rN (approved):

Issue #111, opened 2024-08-27
Re-evaluate WMP fixes


Root-Core:
Perhaps we should re-evaluate all game fixes that involve WMP.
As @layercak3 wrote in https://github.com/Open-Wine-Components/umu-protonfixes/issues/82#issuecomment-2261525964, there have been changes to `winegstreamer` that fix issues with WMP.

Currently used by 19 fixes:

- [ ] 12210
- [ ] 33990
- [ ] 39690
- [ ] 65610
- [ ] 200490
- [ ] 204450
- [ ] 237890
- [ ] 256330
- [ ] 260130
- [ ] 460120
- [ ] 638160
- [ ] 740550
- [ ] 888790
- [ ] 1113000
- [ ] 1144400
- [ ] 1230140
- [ ] 1277930
- [ ] 1829980
- [ ] 2458530

And additionally, there are some fixes using `wmp9_x86_64`, which isn't in `winetricks`.

- [ ] 211420
- [ ] 388750
- [ ] 1097880

R1kaB3rN:
Just tested 1144400 (Senren * Banka) against GE-Proton9-11 and can confirm media playback works. I'm pretty confident that the result will be the same for the other Yuzusoft titles as well.

Root-Core:
CC @CreeperofDoom, since you own the YUZUSOFT games and worked on the fixes.

--------------

The ArcaniA games are a bit iffy.. that might get fixed upstream though.
CoJ:G has no sound, just like Bully. And it can be fixed the same way.. #113.

CreeperofDoom:
> CC @CreeperofDoom, since you own the YUZUSOFT games and worked on the fixes.
> 
> The ArcaniA games are a bit iffy.. that might get fixed upstream though. CoJ:G has no sound, just like Bully. And it can be fixed the same way.. #113.

I only own Yuzusoft’s Riddle Joker through Steam, one title also had the Same Protonfix which i don’t have through Steam. The WMP Video Playback in Riddle Joker doesn’t work and will simply skip playback without the fix on GE-Proton 9-11.

layercak3:
Please attempt with media converter disabled for reasons described in [here](https://github.com/ValveSoftware/Proton/issues/1464#issuecomment-2266666661) (`disable_protonmediaconverter()` or set variables), and if possible a 100% CG save in case of different codecs being used. They work with it disabled, except I didn't test 2458530.
[...]

PR #115, opened 2024-08-29
Update fixes for YUZUSOFT titles

R1kaB3rN:
Related to https://github.com/Open-Wine-Components/umu-protonfixes/issues/111


R1kaB3rN:
CC @layercak3.

A review would be appreciated.

CreeperofDoom:
The new fix works fine, but...

I tested the title Riddle Joker:
One user reported there is a Pixelation issue with the WMP Video Files which i dont know how to fix, applying the vomstyle launch command breaks Video Playback.

The old one with the WMP stuff worked perfect it seemed. But Upgrading GE-Proton would break it, requiring the Proton Data to be cleared to remedy it.

Performance Issues are caused by a broken Pre-Compiled Shader Cache. Deleting that and booting Riddle Joker offline fixes the problem (going online on Steam will force a redownload of the broken cache, Coming at 5 Gigabytes)

R1kaB3rN:
If it works, that's fine. Compared to the old fix, this new one would be preferred as it avoids having new users from manually clearing Proton-related data for future wine prefix upgrades. The `wmp11` verb will need to be thoroughly investigated and resolved separately at another time. 

We run winetricks verbs to fix games, and should be perceived as temporary workarounds until there's a solution upstream. Therefore, if a decent native solution (e.g., gstreamer bad plugins, winegstreamer) exists, it should always be preferred over dubious binaries sourced from winetricks verbs. 

As for potential regressions from wine that may effect playback, that's something everyone who uses GE-Proton should be prepared for and what's certain is that video playback will be broken after the wine prefix upgrade. If there happens to be a regression in winegstreamer, as always, the problem should be reported upstream and we can always reintroduce the old fix in the future when necessary.

CreeperofDoom:
IMG_5658
I get this when attempting to play one of the Video Files in Riddle Joker. GE-Proton 9-13 used

CreeperofDoom:
Actually no Upgrading isn’t the cause. I tried clearing the Proton Data and it didn’t fix it. I have no clue whats causing it because the command didn’t change. It worked prior with the new fix

PR #116, opened 2024-08-29
star citizen gamefixes

mactan-sc:
contingent upon winetricks being of a recent enough version to contain powershell verbs 
https://github.com/Winetricks/winetricks/pull/2256
https://github.com/Winetricks/winetricks/pull/2211

### DIFF
--- a/gamefixes-umu/umu-starcitizen.py
+++ b/gamefixes-umu/umu-starcitizen.py
@@ -1,9 +1,9 @@
 """ Game fix for Star Citizen
 """
-#pylint: disable=C0103
+#pylint: disable=C0103,E0401
 
 import os
-from protonfixes import util # pylint: disable=E0401
+from protonfixes import util
 from protonfixes.logger import log
 
 def main():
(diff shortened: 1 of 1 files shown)


R1kaB3rN:
Alternatively, we can just bisect the relevant patches and create our own custom powershell verb for now, which avoids having to bump winetricks to the latest git.

ProjectSynchro:
Figured I'd chime in:

In terms of why this change is needed, here's some context: https://github.com/lutris/lutris/issues/5577
Essentially, with the forced update to the new 2.0.x launcher, powershell becomes required for basic launcher functionality.

Game data verification, game folder creation and some kind of dependency check all require powershell now (as a dependency).

Currently, The folders that would be created from the launcher itself are made via the protonfix. 

In the 1.6.10 launcher you can get around the powershell requirement by disabling it via a wine dll override and creating the folders yourself. 

Now in the new 2.0.x launcher when verifying the game files (which happens during the process of installing a patch) the launcher falls over and fails due to the fact it checks the status of the game folders.

Unfortunately (or fortunately) it looks like PowerShell is being used to call CMD to run a utility batch file. On Windows you can use that as a method for elevating a process which CIG does, in Wine we get a stub instead.

Some paths forward:
1.  Pull in the verbs I have in the upstream project that is in winetricks here: https://github.com/ProjectSynchro/powershell-wrapper-for-wine/tree/master/verbs

One potential issue with this is the fact that you need to call on another custom verb, which in my experience doesn't seem to always work. (winetricks has no idea the other custom verb exists)

2.  Cherrypick these winetricks commits:
https://github.com/Winetricks/winetricks/commit/07fa5dc819cbab3f4ec43d76767775cc63b61da8
https://github.com/Winetricks/winetricks/commit/d486bc92147b769afa5693890f6031c61eae054f
https://github.com/Winetricks/winetricks/commit/eeaed1f4651a4b19d56891e588a881936ea7a159
https://github.com/Winetricks/winetricks/commit/72b934e1e10c041ec6986f5f2fb4f143d8f6b941

PR #120, opened 2024-09-04
Steam: Fix audio for BLUE REFLECTION (658260)

salixor:
Disable media converter for BLUE REFLECTION, like many other Gust games.


R1kaB3rN:
https://github.com/ValveSoftware/Proton/issues/1522

PR #124, opened 2024-09-08
Added README.md with documentation

### LINKED ISSUE #122: Add Documentation
See https://github.com/Open-Wine-Components/umu-protonfixes/issues/50#issuecomment-2329988597

Root-Core:
I may have missed a few things, but I think this is a good starting point. I'm not a native speaker, so there might be some rough edges.

Feel free to improve and expand.

I also fixed some minor issues I encountered while writing the documentation.

Closes #122

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ We do enforce some linting and testing that can and should be done locally - bef
 To install the necessary tools, just run:
 
 ```bash
-python3 -m pip install --upgrade pip
+python -m pip install --upgrade pip
 pip install ruff
 ```
 
@@ -133,4 +133,17 @@ You can still run it though:
 ```bash
 cd .github/scripts
 python check_gamefixes.py
+```
+
+You need the following prerequisites:
+
+```bash
+python -m pip install --upgrade pip
+pip install ijson
+```
+
+On Arch / Manjaro:
+
+```bash
+sudo pacman -Sy python-ijson
 ```
--- a/README.md
+++ b/README.md
@@ -124,6 +124,12 @@ ruff check .
 
 You are good to go when you see the message `All checks passed!`.
 
+If not, try letting `ruff` (the linter we use) fix it automatically:
+
+```bash
+ruff format .
+```
+
 ### Testing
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -0,0 +1,149 @@
...
+
+https://github.com/Open-Wine-Components/umu-database
+
+## Config file
+
+There is a config file located at:
+
+> ~/.config/protonfixes/config.ini
+
+It's not widely used at the moment, but you can configure some aspects.
```
R1kaB3rN, line 55: Agreed. I doubt this feature is widely used too, so we should definitely consider deprecating this feature along with the GDrive functionality.

README.md
```diff
@@ -0,0 +1,149 @@
...
+On Arch / Manjaro, you should run this instead:
+
+```bash
+sudo pacman -Sy ruff
+```
+
+To run a check, just execute:
+
+```bash
+ruff check .
```
R1kaB3rN, line 122: `ruff format .` will format according to the tables in `ruff.toml`, and should correct any format-related errors (e.g., mixes of spaces/tabs) automatically for most cases.

R1kaB3rN:
Yeah, it's a good starting point. Thanks.

Once I write something up, I'll try to merge this.

README.md
```diff
@@ -0,0 +1,149 @@
...
+```
+
+You are good to go when you see the message `All checks passed!`.
+
+### Testing
+
+The filenames of the fixes are checked against the Steam and GOG APIs. All symbolic links are also checked. This is not mandatory to run locally, as you should have a working game fix to begin with, and it will be done automatically by Github's CI.
```
R1kaB3rN, line 129: Once we get more fixes related to other stores, it'd be nice to validate their store IDs against an API too. Or if there isn't an API, maybe query the game's store page if it includes the game's identifier in the URL like Steam's.

Root-Core: Sure, I was just documenting the current state.

Loosely related, but I had some discussion about the other stores here: https://github.com/Open-Wine-Components/umu-database/issues/39

README.md
Root-Core: I can add that command or change the current one, if you want.

README.md
Root-Core: Or we can add more config parameters and actually use that feature. Not sure yet.

PR #131, opened 2024-09-19
Delete fix for Sleeping Dogs: DE

### LINKED ISSUE #104: Re-evaluate Sleeping Dogs DE fix
https://github.com/Open-Wine-Components/umu-protonfixes/blob/master/gamefixes-steam/307690.py

The fix is probably obsolete due to this commit:

https://github.com/ValveSoftware/wine/commit/85130d1f4cbf7b5d71499d70f76be1ef19afb515

Needs to be re-tested.

Tiagoquix:
Proton 9.0.3 has been released:

https://github.com/ValveSoftware/Proton/releases/tag/proton-9.0-3

> Now playable: 
> - Sleeping Dogs: Definitive Edition

This closes #104.


Root-Core:
Have you tested it actually works? Sometimes things are only working for the SteamDeck.

R1kaB3rN:
I think this is fine to merge. The fix went through a testing phase that takes weeks then was merged and released as stable in 9.0-3., so we should trust it like we're already doing with their other fixes (e.g., CPU core limits). 

If it doesn't work in non-Steam Deck then that means there was a regression, and I think they'll take that seriously when reported. Like always, we can always reintroduce the fix too.

PR #132, opened 2024-09-19
star citizen gamefixes

mactan-sc:
- AMD environment variable is no longer needed

- Nvidia variable enables Vulkan but prevents DLSS. Alternate method of patching libcuda has become favorable as it enables both Vulkan and DLSS

https://github.com/jp7677/dxvk-nvapi/issues/174#issuecomment-2227462795

https://github.com/starcitizen-lug/knowledge-base/wiki/Troubleshooting#dlssdeep-learning-super-sampling--vulkan

### DIFF
--- a/gamefixes-umu/umu-starcitizen.py
+++ b/gamefixes-umu/umu-starcitizen.py
@@ -8,11 +8,5 @@ def main() -> None:
     # eac workaround
     util.set_environment('EOS_USE_ANTICHEATCLIENTNULL', '1')
 
-    # needed for nvidia vulkan
-    util.set_environment('WINE_HIDE_NVIDIA_GPU', '1')
-
-    # needed for amd vulkan
-    util.set_environment('dual_color_blend_by_location', 'true')
-
-    # allow the RSI Launcher to auto-update itself
+    # RSI Launcher depends on powershell
     util.protontricks('powershell')


R1kaB3rN:
In the NVIDIA case, supposing that fix works, have you already explored being able to automate the patching of libcuda.so in a fix so users don't need to do it manually? I think we should strive to do that here.

For instance, what we can do is create another copy, patch it with those instructions, then mount that copy at the same install path. Or copy the file to /tmp, patch it, then set LD_PRELOAD to the patched path. Certainly, that would be better practice than instructing users to patch that file directly in the system path. In fact, the comment says to copy the .so file elsewhere.

> Just copy libcuda.so to some folder and run the command echo -ne $(od -An -tx1 -v libcuda.so | tr -d '\n' | sed -e 's/00 00 00 f8 ff 00 00 00/00 00 00 f8 ff ff 00 00/g' -e 's/ /\\x/g') > libcuda.patched.so... a libcuda.patched.so will be created, then... use LD_PRELOAD env to load libcuda.patched.so( like... LD_PRELOAD=/path/to/the/libcuda.patched.so ).

mactan-sc:
right the star citizen LUG's recommendation presently is to copy the .so and use the LD_PRELOAD env var to use the modified file. Would those sort of file operations be welcome in a protonfix or is that best left to the end user to take action?

R1kaB3rN:
Closing as https://github.com/Open-Wine-Components/umu-protonfixes/pull/136 supersedes this

PR #133, opened 2024-09-20
Environment override to allow God of War: Ragnarok to run on desktop Linux

UsernamesAreNotMyThing:
Game won't start on non-Steam Deck systems without `SteamDeck=1` environment variable, and is otherwise considered to be fully playable with that variable set.

Source: https://github.com/ValveSoftware/Proton/issues/8107#issuecomment-2361421991

### DIFF
--- a/gamefixes-steam/2322010.py
+++ b/gamefixes-steam/2322010.py
@@ -1,0 +1,6 @@
+"Game fix for God of War: Ragnarok"
+from os import environ
+
+def main():
+	# Game won't launch on non-Steam Deck systems without this environment variable
+	environ['SteamDeck'] = '1'


R1kaB3rN:
I think this needs to be fixed upstream by the devs or allowed a bit of time until Valve steps in. The game just released, and a report was just made for this game yesterday. Even if we do set that environment variable to get the game launching, it's insufficient as, reportedly, setting that environment variable causes another problem -- mice look not working.

UsernamesAreNotMyThing:
Looking at reviews on the store page, it looks like the mouse thing might be a game bug.

UsernamesAreNotMyThing:
Also worth noting, even if the original problem is fixed on Proton's end, it's possible that the game may *still* not launch without the variable, due to problems with PSSDK failing with a wine message in the past (namely Ghost of Tsushima multiplayer and Concord, the latter of which is now defunct), which would crash the game. Apparently setting that variable will bypass this behavior and will allow the game to launch anyway, according to the linked thread.

GloriousEggroll:
merging for now, we can always remove later.

PR #136, opened 2024-09-20
Add patch for libcuda.so and implement fix for Star Citizen

ProjectSynchro:
This PR adds a function (patch_libcuda) into util.py which patches libcuda.so with the approach seen here: https://github.com/jp7677/dxvk-nvapi/issues/174#issuecomment-2227462795

This may also be useful for other games that over-allocate memory when enabling DLSS (I know this affects Star Citizen but I don't have NVIDIA so I am not aware of other titles that have this issue.)

Additionally this expands https://github.com/Open-Wine-Components/umu-protonfixes/pull/132 by integrating that into the protonfix.

Please let me know if you'd like any changes. Testers are also welcome, I don't have an NVIDIA GPU to test this fix.. but in theory it should work based on my eyeballing of the code.

This could also be moved into `umu-starcitizen.py` but there may be other affected games this could help with.

### DIFF
--- a/util.py
+++ b/util.py
@@ -454,7 +454,7 @@ def patch_libcuda() -> bool:
         try:
             with open(libcuda_path, 'rb') as f:
                 binary_data = f.read()
-        except IOError as e:
+        except OSError as e:
             log.error(f'Unable to read libcuda.so: {e}')
             return False
 
@@ -465,12 +465,12 @@ def patch_libcuda() -> bool:
         try:
             with open(patched_library, 'wb') as f:
                 f.write(patched_binary_data)
-        except IOError as e:
+        except OSError as e:
             log.error(f'Unable to write patched libcuda.so to {patched_library}: {e}')
             return False
 
         log.info(f'Patched libcuda.so saved to: {patched_library}')
-        
+
         log.info(f'Setting LD_PRELOAD to: {patched_library}')
         set_environment('LD_PRELOAD', patched_library)
         return True
--- a/util.py
+++ b/util.py
@@ -1,6 +1,7 @@
 """Utilities to make gamefixes easier"""
 
 import configparser
+import ctypes.util
 from io import TextIOWrapper
 import os
 import sys
@@ -417,6 +418,66 @@ def winedll_override(dll: str, dtype: Literal['n', 'b', 'n,b', 'b,n', '']) -> No
     protonmain.append_to_env_str(
         protonmain.g_session.env, 'WINEDLLOVERRIDES', setting, ';'
     )
+
+
+def patch_libcuda() -> bool:
+    """Patches libcuda to work around games that crash with DLSS due to wonky memory allocation.
+
+    The patched library is overwritten at every launch and is placed in .cache
+
+    Returns true if the library was patched correctly. Otherwise returns false
+    """
+    cache_dir = os.path.expanduser('~/.cache/protonfixes')
+    os.makedirs(cache_dir, exist_ok=True)
+
+    try:
+        library_base = ctypes.util.find_library('cuda')
+        if not library_base:
+            log.warn('libcuda.so not found on the system.')
+            return False
+
+        library_paths = ['/lib64', '/usr/lib64', '/usr/local/lib64']
+
+        libcuda_path = None
+        for lib_dir in library_paths:
+            full_path = os.path.join(lib_dir, library_base)
+            if os.path.exists(full_path):
+                libcuda_path = os.path.abspath(full_path)
+                break
+
+        if not libcuda_path:
+            log.error(f'libcuda.so found as {library_base}, but not found in standard library directories.')
+            return False
+        log.info(f'Found libcuda.so at: {libcuda_path}')
+
+        patched_library = os.path.join(cache_dir, 'libcuda.patched.so')
+        try:
+            with open(libcuda_path, 'rb') as f:
+                binary_data = f.read()
+        except IOError as e:
+            log.error(f'Unable to read libcuda.so: {e}')
+            return False
+
+        hex_data = binary_data.hex()
+        hex_data = hex_data.replace('000000f8ff000000', '000000f8ffff0000')
+        patched_binary_data = bytes.fromhex(hex_data)
+
+        try:
+            with open(patched_library, 'wb') as f:
+                f.write(patched_binary_data)
+        except IOError as e:
+            log.error(f'Unable to write patched libcuda.so to {patched_library}: {e}')
+            return False
+
+        log.info(f'Patched libcuda.so saved to: {patched_library}')
+        
+        log.info(f'Setting LD_PRELOAD to: {patched_library}')
+        set_environment('LD_PRELOAD', patched_library)
+        return True
+
+    except Exception as e:
+        log.error(f'Unexpected error occurred: {e}')
+        return False
 
 
 def disable_nvapi() -> None:
--- a/util.py
+++ b/util.py
@@ -438,12 +438,11 @@ def patch_libcuda() -> bool:
             log.warn('ldconfig not found in PATH.')
             return False
 
-        # Use subprocess.run with explicit encoding handling
+        # Use subprocess.run with capture_output and explicit encoding handling
         try:
             result = subprocess.run(
                 [ldconfig_path, '-p'],
-                stdout=subprocess.PIPE,
-                stderr=subprocess.PIPE,
+                capture_output=True,
                 check=True
             )
             # Decode the output using utf-8 with fallback to locale preferred encoding
--- a/util.py
+++ b/util.py
@@ -432,9 +432,27 @@ def patch_libcuda() -> bool:
     os.makedirs(cache_dir, exist_ok=True)
 
     try:
-        # Use ldconfig -p to search library paths
+        # Use shutil.which to find ldconfig binary
+        ldconfig_path = shutil.which('ldconfig')
+        if not ldconfig_path:
+            log.warn('ldconfig not found in PATH.')
+            return False
+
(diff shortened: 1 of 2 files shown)


util.py
```diff
@@ -419,6 +421,93 @@ def winedll_override(dll: str, dtype: Literal['n', 'b', 'n,b', 'b,n', '']) -> No
...
+    try:
+        library_base = ctypes.util.find_library('cuda')
+        if not library_base:
+            log.warn('libcuda.so not found on the system.')
+            return False
+
+        library_paths = [
+            '/lib64', '/usr/lib64', '/usr/local/lib64',
+            '/lib', '/usr/lib', '/usr/local/lib'
+        ]
```
R1kaB3rN, line 443: We actually can simplify this a little bit, depending on your goal.

Proton is intended to be use within the Steam Runtime, and the base platform is Debian-based where it's shared library paths can be known ahead of time -- `/usr/lib/x86_64-linux-gnu` and `/usr/lib/i386-linux-gnu`. This was true in the past (steamrt2 AKA soldier) and will be in the future with steamrt4 and steamrt5. 

But you can't find this file there, since it's part of the graphics driver stack. The graphics stack that's mounted by pressure-vessel will be in `/usr/lib/pressure-vessel/overrides/` so you will need to find the file there. The `proton` file makes this simple by setting  `LD_LIBRARY_PATH=LD_LIBRARY_PATH=/usr/lib/pressure-vessel/overrides/lib/x86_64-linux-gnu/aliases:/usr/lib/pressure-vessel/overrides/lib/i386-linux-gnu/aliases`. Just be sure to move up from the `aliases` directory.

R1kaB3rN: But if your goal is to also support Proton not being run inside its runtime, parse the output `ldconfig -p` or `ldconfig -XNv` for the path candidates. The Python standard library actually does the former and this approach would be a more general solution than the above.

R1kaB3rN: Also, if you're interested to see the files in the Steam Runtime container for debugging purposes, see:

https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/-/blob/main/docs/slr-for-game-developers.md?ref_type=heads#running-in-an-interactive-shell

util.py
```diff
@@ -419,6 +421,93 @@ def winedll_override(dll: str, dtype: Literal['n', 'b', 'n,b', 'b,n', '']) -> No
     )
 
 
+def patch_libcuda() -> bool:
+    """Patches libcuda to work around games that crash with DLSS due to wonky memory allocation.
```
R1kaB3rN, line 425: I'm fine having this function because I think it would be useful. Hopefully this fix is short lived and it gets properly addressed by the CUDA engineers or some other project upstream.

I'm interested in more context and the scale of this problem though. For instance, does this occur in Windows as well and does this effect every game when DLSS is enabled?

util.py
```diff
@@ -419,6 +421,93 @@ def winedll_override(dll: str, dtype: Literal['n', 'b', 'n,b', 'b,n', '']) -> No
...
+        try:
+            with open(libcuda_path, 'rb') as f:
+                binary_data = f.read()
+        except OSError as e:
+            log.error(f'Unable to read libcuda.so: {e}')
+            return False
+
+        hex_data = binary_data.hex()
+        hex_data = hex_data.replace('000000f8ff000000', '000000f8ffff0000')
+        patched_binary_data = bytes.fromhex(hex_data)
```
R1kaB3rN, line 484: I think we need a comment and a link so this looks less suspect

R1kaB3rN (changes requested):

util.py
ProjectSynchro: I see.. I'll have to rework this a bit. I was initially intending to avoid anything non-python for portability's sake, but calling ldconfig would work too as cutils does that as well like you said. My initial attempt was to avoid anything outside of python as a bit of a challenge, but parsing the output of ldconfig simplifies this a fair bit. 

My mind is still stuck in the days of pre-pressure vessel proton.. whoops. Tthe Linux runtime never occurred to me, I'll poke around in a shell to check on locations of the library.

util.py
ProjectSynchro: I'm not too well versed in CUDA, let alone the nvidia driver stack, but as far as I can tell from a distance this seems to be a difference in the approach to memory allocation in Linux versus Windows. 

> When creating a Vulkan device with VK_NVX_binary_import and/or VK_NVX_image_view_handle extensions... libcuda tries to allocate memory in a specific area already used by the game, so... failing.
DXVK and VKD3D get around this by recreating the device without these extensions... but without these extensions DLSS does not work.
See: https://github.com/jp7677/dxvk-nvapi/issues/174#issuecomment-2227462795 onwards for additional context (seems like a conversation between an NVIDIA employee/contractor and the individual who found a workaround to the issue.)

This could be rearing it's head due to a Linux specific platform issue (with those two above extensions) or it could also be that the CUDA stack on Windows protects itself better against issues like this, probably due to the size/ demographics of the userbase.

Anyhow hope that gives you some context as to why this is needed for Star Citizen at the moment (and potentially others if this implementation issue is repeated in the future)

util.py
ProjectSynchro: Indeed, I was going for minimal comments, but without context this definitely can look sketchy. I'll change this.

ProjectSynchro: Should be fixed with https://github.com/Open-Wine-Components/umu-protonfixes/pull/136/commits/09785517d12838d0789da6c9c385714b6703f56b

let me know if you want more (or less) information

util.py
ProjectSynchro: I used ldconfig -p's output to find the correct library. This is nice as ldconfig will also tell us which arch the library is (hence the check for the x86-64 string) which allows me to remove the need for checking if the ELF file is indeed a 64bit binary or not.

https://github.com/Open-Wine-Components/umu-protonfixes/pull/136/commits/8a27ea21d16aac2ba4cf2734450edd07b70c6b47 fixes this, and should pull in the correct paths as it searches overrides as well since `ld.so.cache` should be updated with libraries within them.

ProjectSynchro:
Should be good to go now, both for use within and outside of the Linux runtime.

util.py
```diff
@@ -419,6 +419,86 @@ def winedll_override(dll: str, dtype: Literal['n', 'b', 'n,b', 'b,n', '']) -> No
...
+
+    Returns true if the library was patched correctly. Otherwise returns false
+    """
+    cache_dir = os.path.expanduser('~/.cache/protonfixes')
+    os.makedirs(cache_dir, exist_ok=True)
+
+    try:
+        # Use ldconfig -p to search library paths
+        try:
+            output = subprocess.check_output(['ldconfig', '-p'], stderr=subprocess.STDOUT, text=True)
```
R1kaB3rN, line 437: Here, I would use either `subprocess.run` or `subprocess.Popen` to create the subprocess  so we can have some guarantee of the text returned from the command. This can be a problem, especially when running outside the Steam Runtime, as character encoding is currently platform/host dependent until Python 3.15.

See https://github.com/python/cpython/blob/main/Lib/ctypes/util.py#L301

R1kaB3rN (changes requested):

util.py
R1kaB3rN: I would maybe specify the absolute path to the executable too by either hardcoding the path or use` shutil.which` to find it. I think it would be acceptable to hardcode `/sbin/ldconfig` as the standard library does it, which signals to me that it's stable and a fundamental tool. Same with the Steam Runtime code, and the Steam client's `run.sh` script at `$HOME/.local/share/Steam/ubuntu12_32/steam-runtime`

ProjectSynchro: This should be resolved with https://github.com/Open-Wine-Components/umu-protonfixes/pull/136/commits/338d9e5fb36625015196a5a9c306d48fc278d7bc

Used `subprocess.Popen` as CI doesn't like `subprocess.run` apparently.

util.py
```diff
@@ -419,6 +419,109 @@ def winedll_override(dll: str, dtype: Literal['n', 'b', 'n,b', 'b,n', '']) -> No
...
+    try:
+        # Use shutil.which to find ldconfig binary
+        ldconfig_path = shutil.which('ldconfig')
+        if not ldconfig_path:
+            log.warn('ldconfig not found in PATH.')
+            return False
+
+        # Use subprocess.Popen with explicit encoding handling
+        try:
+            process = subprocess.Popen(
```
R1kaB3rN, line 443: I would follow the linter's recommendation instead of working around it. Using `capture_output` simplifies things a bit and as a side effect it would also avoid the subprocess's resources (e.g., file descriptors) potentially not all being released if the subprocess's execution was interrupted.

See https://docs.astral.sh/ruff/rules/replace-stdout-stderr/

R1kaB3rN (changes requested):

util.py
R1kaB3rN: Either would be fine:

`subprocess.run([ldconfig_path, '-p'], capture_output=True).output`

or

```
with Popen([ldconfig_path, '-p'], stdout=subprocess.PIPE, stderr=subprocess.PIPE):
  ...
```

R1kaB3rN:
Overall, this looks ok to me.

@mactan-sc for testing or someone else as unfortunately I do not have a NVIDIA GPU in my system.

util.py
ProjectSynchro: Solved with https://github.com/Open-Wine-Components/umu-protonfixes/pull/136/commits/17a428ec8030f57e3c26f42e9b04e1e5acf290a0

R1kaB3rN (approved):

the-sane:
FYI: We've put out a request for testing to our community on the Star Citizen LUG discord. We'll report back with the results.

PR #140, opened 2024-09-22
Game fix for Plain Sight

iodream:
Plain Sight needs XNA 3.1 or it will get stuck when entering any level. 

I've noticed executing the verb xna31 works with `winetricks-next` but can fail with currently stable version 20240105.  I guess if it should fail some time in the future it would just fail silently in the background and game would launch regardless with Proton just without the fix.    

https://github.com/ValveSoftware/Proton/issues/7594

### DIFF
--- a/gamefixes-steam/49900.py
+++ b/gamefixes-steam/49900.py
@@ -1,0 +1,7 @@
+"""Game fix for Plain Sight"""
+
+from protonfixes import util
+
+def main() -> None:
+    """Installs XNA 3.1 to avoid crash when loading levels"""
+    util.protontricks('xna31')


iodream:
Not sure why the CI fails, i ran the same test locally and it passed

Root-Core:
I think that Steam blocked the requests or similar. That has nothing to do with your changes, so it won't block the merge.

R1kaB3rN:
The Steam App ID 12210 is GTA IV, and judging from the store page it seems like it isn't for sale anymore.

https://steamdb.info/app/12210/

https://store.steampowered.com/app/901583/Grand_Theft_Auto_IV_Complete_Edition/

R1kaB3rN:
> Not sure why the CI fails, i ran the same test locally and it passed

I'm aware of this issue and to solve it, we'd need to find another public API to validate against.

Root-Core:
> I'm aware of this issue and to solve it, we'd need to find another public API to validate against.

We could try querying SteamDB with all failed IDs. This could at least reduce the risk of missing IDs.

My idea would be to only check the changed / added IDs though.. we have verified all current IDs. Is there a hook for getting changed / added files, that we could use as an argument for the script?

PR #141, opened 2024-09-23
Added fix for Gothic Playable Teaser

Root-Core:
The game is unplayable slow without these settings.
It also creates gigabytes of logs, which this fix cleans up.

### DIFF
--- a/gamefixes-steam/950670.py
+++ b/gamefixes-steam/950670.py
@@ -1,0 +1,53 @@
+"""Game fix for Gothic Playable Teaser"""
+
+from pathlib import Path
+from protonfixes import util
+from protonfixes.logger import log
+
+
+def main() -> None:
+    """Performance fixes and cleanup of excessive logs"""
+    modify_settings()
+    clear_logs()
+
+
+def modify_settings() -> None:
+    # Assure that the settings folder exists
+    path = Path('GothicRemake/Saved/Config/WindowsNoEditor/')
+    if not path.is_dir():
+        log(f'Creating settings folder "{path}".')
+        path.mkdir(parents=True, exist_ok=True)
+
+    # Assure, that the settings file exists
+    # Necessary defaults will still be created by the game
+    path = path / 'Engine.ini'
+    if not path.is_file():
+        log(f'Creating empty settings file "{path}".')
+        path.touch(exist_ok=True)
+
+    # Disable excessive (as in gigabytes) logging
+    # Disable motion blur / depth of field / lens flare
+    game_opts = """
+    [Core.Log]
+    Global=off
+    LogInteractiveProcess=all off
+
+    [SystemSettings]
+    r.MotionBlur.Max=0
+    r.MotionBlurQuality=0
+    r.DefaultFeature.MotionBlur=0
+    r.DepthOfFieldQuality=0
+    r.LensFlareQuality=0
+    """
+    util.set_ini_options(game_opts, path, 'utf-8', 'game')
+
+
+def clear_logs() -> None:
+    # Clear all logs > 10 MB
+    log_files = Path('GothicRemake/Saved/Logs').glob('*.log')
+    for file in log_files:
+        file_size = file.stat().st_size         # Bytes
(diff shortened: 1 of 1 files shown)


gamefixes-steam/950670.py
```diff
@@ -0,0 +1,53 @@
...
+
+def main() -> None:
+    """Performance fixes and cleanup of excessive logs"""
+    modify_settings()
+    clear_logs()
+
+
+def modify_settings() -> None:
+    # Assure that the settings folder exists
+    path = Path('GothicRemake/Saved/Config/WindowsNoEditor/')
```
R1kaB3rN, line 16: For sanity, I would use `util.get_game_install_path` as `STEAM_COMPAT_INSTALL_PATH` is a Steam property so we can have _some_ guarantees of the path and the files that will be removed. I know the Steam client changes to the base game's directory before running the game, but, unfortunately, some users run GE-Proton outside Steam _and_ its runtime, so we might not even be in the expected directory. At worst, this fix would be creating/removing files in their home directory.

gamefixes-steam/950670.py
```diff
@@ -0,0 +1,53 @@
...
+
+from pathlib import Path
+from protonfixes import util
+from protonfixes.logger import log
+
+
+def main() -> None:
+    """Performance fixes and cleanup of excessive logs"""
+    modify_settings()
+    clear_logs()
```
R1kaB3rN, line 11: Hard to say if we should do this. Clearing the logs would be a plus, but wouldn't make the game more playable compared to the `modify_settings`.

R1kaB3rN: Personally, though, I would want the logs removed to save some space. 1GB+ of generated logs would be a problem for me and I'd hate it.

R1kaB3rN:
Also, this game appears to have a 64-bit executable at `GothicRemake/Binaries/Win64/GothicRemake.exe` and Steam doesn't appear to have a launch option to use it like other titles. Not sure why the file is significantly larger though. Have you already running it?

See https://steamdb.info/depot/950671/

gamefixes-steam/950670.py
Root-Core: My thinking was to "fix" what the game was doing wrong. So it could be seen as a kind of "game fix." 😊 

It's pretty unlikely to cause any harm, but it might also be seen as outside the scope of what we're aiming for.

Root-Core:
I will check out the 64bit executable when I find the time. Looks interesting.

gamefixes-steam/950670.py
Root-Core: Shouldn't we make sure we are in the right directory before running the fix?
For example, by checking and possibly correcting it in `fix.py`?

This would make all fixes cleaner, easier to read and implement.

R1kaB3rN: I think we should continue depending on the client to change to the correct directory and in the case of Steam, it's `STEAM_COMPAT_INSTALL_PATH`. Only the developers can know with certainty what the correct directories are for the game and they typically make it available to the client in some way (e.g., environment variables, configuration files, etc.). I think `protonfixes` shouldn't attempt to change directories or attempt to find the correct one, unless the developer had failed to inform the client that metadata somehow or if we want to enforce running the 64-bit executable, which typically require to change to the subdirectory of the base game's directory.  Therefore, if a client like Lutris or Heroic fails to change to the correct directory, it's on them.

R1kaB3rN: In the context of this fix, I think it would be better to use `util.get_game_install_path` as it would avoid hardcoding any paths specific to the game and support non-Steam clients as they would be able to override `STEAR_COMPAT_INSTALL_PATH`.

R1kaB3rN: Hmm, but yeah. Maybe there should be a prerequisite before applying any fixes if some are going to be manipulating or creating files. Perhaps we can skip applying Steam fixes all together if `STEAR_COMPAT_INSTALL_PATH` isn't set in the environment...?

Root-Core:
The 64-bit executable is actually used, according to btop.

PR #145, opened 2024-10-05
star citizen gamefixes

mactan-sc:
- Remove EAC workaround
  -  Enforcement is being enabled in Star Citizen 3.24.2
  -  A solution for EAC has been identified, requiring user action in the RSI Launcher to change game files to a location on `Z:\` instead of `C:\`
- ~~Add `arial` and `tahoma` fonts that are missing for some users~~
- `libglesv2` set to `builtin` improves RSI Launcher animation responsiveness

### DIFF
--- a/gamefixes-umu/umu-starcitizen.py
+++ b/gamefixes-umu/umu-starcitizen.py
@@ -4,13 +4,12 @@ from protonfixes import util
 
 
 def main() -> None:
-    """EAC Workaround"""
-    # eac workaround
-    util.set_environment('EOS_USE_ANTICHEATCLIENTNULL', '1')
-
     # patch libcuda to workaround crashes related to DLSS
     # See: https://github.com/jp7677/dxvk-nvapi/issues/174#issuecomment-2227462795
     util.patch_libcuda()
 
     # RSI Launcher depends on powershell
     util.protontricks('powershell')
+
+    # RSI Launcher animation
+    util.winedll_override("libglesv2", "builtin")


gamefixes-umu/umu-starcitizen.py
```diff
@@ -4,13 +4,16 @@
...
-
     # patch libcuda to workaround crashes related to DLSS
     # See: https://github.com/jp7677/dxvk-nvapi/issues/174#issuecomment-2227462795
     util.patch_libcuda()
 
     # RSI Launcher depends on powershell
     util.protontricks('powershell')
+
+    # RSI Launcher fonts
+    util.protontricks('arial')
```
R1kaB3rN, line 15: Do you mind elaborating a bit why the fake Arial font bundled in Proton is insufficient and how it can be missing?

See:
- https://github.com/GloriousEggroll/proton-ge-custom/blob/master/Makefile.in#L1062
- https://github.com/Open-Wine-Components/umu-proton/blob/proton_9.0/Makefile.in#L811
- https://github.com/ValveSoftware/Proton/blob/proton_9.0/Makefile.in#L810

R1kaB3rN: I ask this because if missing fonts only happens for only a subset of users then I don't think it would be fair to enforce installing it for everyone as some font installations can be 100MB+

R1kaB3rN: In addition to the size increase of the prefix, there's also the download and installation time of the winetricks verb. Some of them take really long, in particular wmp11 verb, so doing more than necessary would be a pretty bad UX

R1kaB3rN: In the worst case,  with `allfonts`, it can 2 min+. 

See this interesting project on how the [installation of the some font-related verbs can be sped up, but with some risks](https://github.com/awaken1ng/winetricksfast).

mactan-sc: the report was from nix users, they were using the release tar.gz of proton-ge. with fonts missing certain inputs and buttons don't always render. if this is too much an edge case given the performance impact the font lines can be removed as well. installers such as lutris includes them already

PR #146, opened 2024-10-06
Remove EAC workaround

DistantThunder:
EAC has been enabled on the servers. It is not possible to play ignoring the AC anymore and so this workaround has become incompatible with current server versions.

* https://github.com/starcitizen-lug/knowledge-base/wiki/Tips-and-Tricks#easy-anti-cheat

### DIFF
--- a/gamefixes-umu/umu-starcitizen.py
+++ b/gamefixes-umu/umu-starcitizen.py
@@ -4,9 +4,7 @@ from protonfixes import util
 
 
 def main() -> None:
-    """EAC Workaround"""
-    # eac workaround
-    util.set_environment('EOS_USE_ANTICHEATCLIENTNULL', '1')
+    """Patches libcuda & installs required dependencies"""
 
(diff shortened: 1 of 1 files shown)


ProjectSynchro:
Just putting in my two cents here. There are a couple other things at play that would have this become a breaking change.

At the moment you need to pass a `Z:/` path into the launcher for EasyAntiCheat to work correctly (as mentioned in the above wiki article)
This essentially means that the game goes from working OOTB to not working without manual tweaks.

Additionally, EAC enforcement isn't currently active on LIVE servers and is in early testing in the PTU.

Currently we are collaborating with CIG staff to find a solution to this that doesn't require manual tweaks to the launcher in order for EAC to function correctly.

ProjectSynchro:
This is also a duplicate of https://github.com/Open-Wine-Components/umu-protonfixes/pull/145

DistantThunder:
Duplicate. Sorry, not seen.

PR #156, opened 2024-10-25
Delete Renegade Ops gamefix

Root-Core:
The missing loop unrolling was implemented in vkd3d. Works with Proton Experimental without the fix.

https://gitlab.winehq.org/wine/vkd3d/-/commit/2034a8bab9acbf3e80d35d597d1565ade3f68077

### DIFF
--- a/gamefixes-steam/99300.py
+++ b/gamefixes-steam/99300.py
@@ -1,8 +1,0 @@
-"""Game fix Renegade Ops"""
-
-from protonfixes import util
-
-
-def main() -> None:
-    """This fixed the black screen issue"""
-    util.protontricks('d3dcompiler_47')


R1kaB3rN (changes requested):
Do you mind referencing the commit in Valve's fork instead of upstream wine where missing loop unrolling was implemented?

Root-Core:
Sure: https://github.com/ValveSoftware/vkd3d/commit/2034a8bab9acbf3e80d35d597d1565ade3f68077
It's the same commit hash.

That's used in https://github.com/ValveSoftware/Proton/tree/experimental_9.0 - as mentioned, the game runs on a fresh prefix in experimental, but not in 9.0. I tested it switching back and forth.

R1kaB3rN (approved):

PR #159, opened 2024-11-14
Add fix for dirt 3 outside of steam

polluxau:
Without this library (openal) dirt 3 wont launch on proton 8 or above outside of steam 

This is for running the game on its own with no store available 

Database pull request
https://github.com/Open-Wine-Components/umu-database/pull/50

### DIFF
--- a/gamefixes-umu/321040.py
+++ b/gamefixes-umu/321040.py
@@ -1,0 +1,1 @@
+/home/pollux/umu-protonfixes-dirt3/gamefixes-steam/321040.py
--- a/gamefixes-umu/321040.py
+++ b/gamefixes-umu/321040.py
@@ -1,1 +1,0 @@
-/home/pollux/umu-protonfixes-dirt3/gamefixes-steam/321040.py
--- a/gamefixes-umu/umu-321040.py
+++ b/gamefixes-umu/umu-321040.py
@@ -1,9 +1,1 @@
-"""Game fix for Dirt 3 Complete Edition"""
-
-
-from protonfixes import util
-
-
-def main() -> None:
-    """Installs OpenAL library, without it the game simply wont launch on proton 8 or above"""
-    util.protontricks('Openal')
+../gamefixes-steam/321040.py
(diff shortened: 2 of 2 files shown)


gamefixes-umu/umu-321040.py
```diff
@@ -0,0 +1,9 @@
+"""Game fix for Dirt 3 Complete Edition"""
+
+
+from protonfixes import util
+
+
+def main() -> None:
+    """Installs OpenAL library, without it the game simply wont launch on proton 8 or above"""
+    util.protontricks('Openal')
```
R1kaB3rN, line 9: Instead of a file, this should be a symbolic link to the corresponding Steam fix

R1kaB3rN (changes requested):

R1kaB3rN:
I won't ask how you acquired this game. But this game is broken under Proton so I'm fine with creating this fix.

polluxau:
> I won't ask how you acquired this game. But this game is broken under Proton and we already had a fix for it, so I'm fine with creating this fix.

Do you want the symbolic link still?

gamefixes-umu/umu-321040.py
R1kaB3rN: This needs to be deleted. Remove the file locally then commit and force push your changes.

R1kaB3rN:
> Question also, if its a symbolic link the file is 321040.py, will it still be detected if its like that on a copy of dirt 3 outside of steam from no store?

Yes, but detection and usage of the fix depends. 

The entire system of applying fixes relies on metadata being available and the user/client giving umu that info. For example, when setting `GAMEID=umu-0` and `STORE=none` or `STORE=`, by default, the module will search the `umu-database.csv` for `umu-0`. On the other hand, if the client (e.g., Lutris) sets the ID and store, then protonfixes may not apply the fix.

gamefixes-umu/321040.py
```diff
@@ -0,0 +1 @@
+/home/pollux/umu-protonfixes-dirt3/gamefixes-steam/321040.py
```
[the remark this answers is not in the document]
polluxau: I see, thank you for the help :)

R1kaB3rN (approved):

PR #166, opened 2024-12-09
Add fix for Bright Memory (955050)

loathingKernel:
Game is also available on GOG but I didn't test that version.


R1kaB3rN:
https://github.com/ValveSoftware/Proton/issues/4384

PR #167, opened 2024-12-10
Disable Esync and Fsync for Supreme Commander

UsernamesAreNotMyThing:
Bad performance unless Esync and Fsync are disabled.

ValveSoftware/Proton#1107


R1kaB3rN:
https://github.com/ValveSoftware/Proton/issues/688

Issue #102, opened 2024-08-10
Additional rationale for UPlay fixes


Tiagoquix:
@R1kaB3rN Hello!

I have additional comments about #94:

I agree with your resolution and I believe we should, then, never disable UPlay overlay by default, no matter which game.

Why?

Because UPlay problems are universal. The launcher itself is always the same: if it works wonderfully in one game, it also works wonderfully in others. If it works horribly in one game, it also works horribly in others. It's the same logic as the Steam Client In-Game Overlay.

With the reasoning that we should always prioritize not losing features in certain games, I believe that we should then allow the user to disable the Ubisoft overlay if they wish, but `protonfixes` should not do that. As I see it, `protonfixes`, then, should only add things that are *necessary* to make the game playable, and not just because it makes them better.

For instance, think about old Rockstar Games. Many of them have a mod called "SilentPatch", to fix various game issues, yet it is not absolutely necessary to play the game. You can still play the game without it, but in a very inferior state.

While in some games the UPlay overlay can cause performance and input problems, the game is still playable even with these problems (even if it's in a degraded state).

To conclude, I suggest we remove any game fix related to disabling the UPlay overlay by default.

The best alternative I can think is to add an env. variable (runtime option) exclusive to Proton-GE called `PROTON_GE_DISABLE_UPLAY_OVERLAY` or something like that to make it easy for users to disable UPlay overlay.

I'm also fine if the env. variable is not added and we just remove the UPlay overlay-related fixes and call it a day.

Thanks for your attention.

PR #169, opened 2024-12-14
Remove deactivation of UPlay overlay

Tiagoquix:
See https://github.com/Open-Wine-Components/umu-protonfixes/issues/102 for context.

Not sure if the `@once` should be kept, but I kept it just in case.

This closes https://github.com/Open-Wine-Components/umu-protonfixes/issues/102.

### DIFF
--- a/gamefixes-steam/359550.py
+++ b/gamefixes-steam/359550.py
@@ -4,6 +4,5 @@ from protonfixes import util
 
 
 def main() -> None:
-    """Rainbow Six Siege needs vk_x11_override_min_image_count=2 for AMD, and overlay disabled for Vulkan"""
-    util.disable_uplay_overlay()
+    """Rainbow Six Siege needs vk_x11_override_min_image_count=2 for AMD"""
     util.set_environment('vk_x11_override_min_image_count', '2')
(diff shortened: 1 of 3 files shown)


Root-Core:
I don't think removing the fix is beneficial as long as it solves a problem.
I agree, we could add a toggle via ENV or config and not rely on a game specific fix. At the moment though, we don't really do engine or launcher-wide patches. Maybe we should.

R1kaB3rN:
I think we should just keep the current fixes for now that disable the uplay overlay, but be strict about allowing future fixes that disable them as we know it can potentially break core in-game functionality. 

What problems are you observing in the game play with AC and Rainbow Six specifically in the context of disabling uplay overlay?

R1kaB3rN:
Since you only own 1/2 of them and cannot observe any problems with the game that this fix introduces, then I'm not seeing what this change gains for us besides the risk of severely degrading the performance for the user.

> I believe that the current fixes are outdated and should be removed. Many improvements have been done to Proton since those fixes were added, so the overlay probably works fine.

In the context of this change, the only thing that matters is how the uplay overlay affects the gaming experience for the user when playing these specific titles. Not with how you feel, believe or suspect that will probably work fine. [And if we're to trust the user ProtonDB reports, then this change will just hurt the UX as future users would have to disable the overlay manually](https://www.protondb.com/app/812140). Unless you're able to demonstrate or provide evidence with logs/upstream issue/commit of how disabling it is detrimental or is now obsolete, I won't accept this change.

PR #172, opened 2024-12-15
Workaround for bad Nvidia performance in Indiana Jones and the Great Circle

UsernamesAreNotMyThing:
This includes known workarounds for graphics driver bugs encountered in-game using RadV and Nvidia:
- RadV: Missing textures and ~~glowing eyes~~ (that was actually a game bug)
- Nvidia: Bad performance due to poor VRAM utilization

amdvlk is not known to have any problems with this game.

www.protondb.com/app/2677660?device=any

### DIFF
--- a/gamefixes-steam/2677660.py
+++ b/gamefixes-steam/2677660.py
@@ -1,0 +1,14 @@
+"""Game fix for Indiana Jones and the Great Circle"""
+
+from protonfixes import util
+
+
+def main() -> None:
+	"""Workarounds for various graphics driver bugs encountered in this game"""
+	# RadV: Missing textures and glowing eyes
+	util.set_environment('radv_legacy_sparse_binding', 'true')
+	util.set_environment('radv_zero_vram', 'true')
+	util.set_environment('RADV_DEBUG', 'nodcc')
+	# Nvidia: bad performance due to poor VRAM utilization
+	util.set_environment('DXVK_NVAPI_GPU_ARCH', 'GA100')
+	util.set_environment('__GL_13ebad', '0x1')
--- a/gamefixes-steam/2677660.py
+++ b/gamefixes-steam/2677660.py
@@ -4,11 +4,6 @@ from protonfixes import util
 
 
 def main() -> None:
-	"""Workarounds for various graphics driver bugs encountered in this game"""
-	# RadV: Missing textures and glowing eyes
-	util.set_environment('radv_legacy_sparse_binding', 'true')
-	util.set_environment('radv_zero_vram', 'true')
-	util.set_environment('RADV_DEBUG', 'nodcc')
-	# Nvidia: bad performance due to poor VRAM utilization
+	"""Bad performance on Nvidia due to poor VRAM utilization"""
 	util.set_environment('DXVK_NVAPI_GPU_ARCH', 'GA100')
 	util.set_environment('__GL_13ebad', '0x1')


gamefixes-steam/2677660.py
```diff
@@ -0,0 +1,14 @@
+"""Game fix for Indiana Jones and the Great Circle"""
+
+from protonfixes import util
+
+
+def main() -> None:
+	"""Workarounds for various graphics driver bugs encountered in this game"""
+	# RadV: Missing textures and glowing eyes
```
R1kaB3rN, line 8: Skimming through the relevant upstream Mesa issue, [the graphics-related issue when using RADV seems to have already been resolved for this game](https://gitlab.freedesktop.org/mesa/mesa/-/issues/12257). Remove the 3 lines below.

cc @matte-schwartz for confirmation

R1kaB3rN (changes requested):

gamefixes-steam/2677660.py
```diff
@@ -0,0 +1,14 @@
...
+	util.set_environment('radv_legacy_sparse_binding', 'true')
+	util.set_environment('radv_zero_vram', 'true')
+	util.set_environment('RADV_DEBUG', 'nodcc')
+	# Nvidia: bad performance due to poor VRAM utilization
```
R1kaB3rN: For the NVIDIA driver bug, where's the relevant upstream issue? This needs to be reported upstream so if there isn't one already, please create one as a ProtonDB report will not suffice.

R1kaB3rN (changes requested):

gamefixes-steam/2677660.py
matte-schwartz: correct, and they will all be in the upcoming Mesa stable release this week

gamefixes-steam/2677660.py
UsernamesAreNotMyThing: https://forums.developer.nvidia.com/t/bug-indiana-jones-and-the-great-circle-only-runs-at-40-of-max-tdp-on-rtx-4090/315916/55

R1kaB3rN: Never mind. Found it:

https://forums.developer.nvidia.com/t/bug-indiana-jones-and-the-great-circle-only-runs-at-40-of-max-tdp-on-rtx-4090/315916/55

gamefixes-steam/2677660.py
```diff
@@ -0,0 +1,9 @@
+"""Game fix for Indiana Jones and the Great Circle"""
+
+from protonfixes import util
+
+
+def main() -> None:
+	"""Bad performance on Nvidia due to poor VRAM utilization"""
+	util.set_environment('DXVK_NVAPI_GPU_ARCH', 'GA100')
```
R1kaB3rN, line 8: Sorry, but does setting __GL_13ebad=0x1 not suffice for this game? What happens if you do not set DXVK_NVAPI_GPU_ARCH=GA100?

R1kaB3rN (changes requested):

gamefixes-steam/2677660.py
R1kaB3rN: According to the OP:

> Please note that the game currently needs the env variable DXVK_NVAPI_GPU_ARCH=GA100 %command% to fake as an Ampere card, in order to avoid a game crash under Proton.

R1kaB3rN (approved):

R1kaB3rN:
https://github.com/ValveSoftware/Proton/issues/8292

PR #180, opened 2024-12-17
Add fix for Bus Simulator 21 Next Stop

BlueInterlude:
Epic link: https://store.epicgames.com/en-US/p/bus-simulator-21

Database PR: https://github.com/Open-Wine-Components/umu-database/pull/57

### DIFF
--- a/gamefixes-steam/976590.py
+++ b/gamefixes-steam/976590.py
@@ -1,0 +1,8 @@
+"""Game fix for Bus Simulator 21 Next Stop"""
+
+from protonfixes import util
+
+
+def main() -> None:
+    # Requires vcrun2022 to launch
+    util.protontricks('vcrun2022')
--- a/gamefixes-amazon/umu-976590.py
+++ b/gamefixes-amazon/umu-976590.py
@@ -1,0 +1,1 @@
+../gamefixes-steam/976590.py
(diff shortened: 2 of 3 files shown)


gamefixes-steam/976590.py
```diff
@@ -0,0 +1,8 @@
+"""Game fix for Bus Simulator 21 Next Stop"""
+
+from protonfixes import util
+
+
+def main() -> None:
+    # Requires vcrun2022 to launch
+    util.protontricks('vcrun2022')
```
R1kaB3rN, line 8: [There isn't an upstream issue at all for this title](https://github.com/ValveSoftware/Proton/issues?q=976590), so a fix shouldn't be needed for the Steam version. If there is an issue for it, please report it.

R1kaB3rN (changes requested):

gamefixes-steam/976590.py
R1kaB3rN: Furthermore this game is Deck Verified and [if we're to trust the ProtonDB user reports, there shouldn't be any issues](https://www.protondb.com/app/976590). So I would prefer to keep the fix local to EGS.

R1kaB3rN (approved):

PR #152, opened 2024-10-17
Static type checking / problem annotation / improvements / refactoring

Root-Core:
This PR enables static type checking and annotation of CI issues on GitHub.

It has not been tested much though, so I am opening it as a draft PR.
I'm also not sure if the relative imports will cause problems.
Feedback and testing welcome.

- Fixed all type errors
- Added enums / type aliases
- Added docstrings
- Removed some ruff rule exceptions
- Use `pathlib` in `util.py` and gamefixes
- Refactoring of some gamefixes
- Added some functions to `util.py`
    - create_dos_device()
    - patch_conf_value()
    - patch_voodoo_conf()
    - get_path_syswow64()
- Reimplemented some functions in `util.py`
    - parsing of Proton version strings
    - get_resolution() with RegEx
    - _get_case_insensitive_name()
    - _killhanging()
    - _checkinstalled()
- Implemented new `ConfigBase` class
- Use config for caching dir
- Refactored `logger.py`
- et cetera..

CC #50 

-----

EDIT:
02/21 commits cherry-picked in #200

1. e3771ff0877c367aa0b0817d2ca3c308d4a89005
2. 71eb0da8e21177c9a08491e4617875633b8f5a6a

10/21 commits cherry-picked in #201

1. fd8b058287c6f239a9e07315f6d2cf2c945a06c1
2. ec2292f6c27e64f3128c57311ee289d3fa6c957a
3. 6d7dc326a312ab5a725100d5570736f389aa73b4
4. ec4a37d925389c5eb04c3e1e4e4d8c9c6c1283da
5. a51a512addf8b03b273e0344484fbb5b4fa3a3ef
6. f49e7fa0beda1037781d6ecba37c24587f9fb6f7
7. ad470d5fbc0a3299d2d5d97ef39e67ae10f0f547
8. 1ce9d45e729e3cb4e4544033f9159c97cc4882a7
9. 54c18c266bc39066a946eae6d746daa345b8d132
10. 5142e5cfeb4322f5122e112cc1cb0a65870fd5cd

04/21 commits cherry-picked in #202

1. 02a02136571176c6d4000ea9989ced153aa4cf86
2. 19f61d016bf5292209164618fbc70424ce2fbc0f
3. d738146d196dd249b21dd9262dd6ff319c408620
4. 611641eb5b2424dbef9f1b18b6e524bc2559f90d
- ad470d5fbc0a3299d2d5d97ef39e67ae10f0f547 (parts as new commit)

02/21 commits cherry-picked in #203

[...]

Issue #50, opened 2024-03-29
TODO: Enforce *.py files via the linter


R1kaB3rN:
Related to https://github.com/Open-Wine-Components/umu-protonfixes/pull/48#issuecomment-2025754406

Currently, only the gamefix modules are linted but it would a good idea to also lint the rest of the Protonfix modules as well for quality assurance and to prevent bugs such as https://github.com/Open-Wine-Components/umu-protonfixes/commit/ac9ddf2660f4cd79ddc157424ee238926a4d6d30

Root-Core:
If I don't misunderstand you, this is already the case with commit c956ac9689e59391c6df3380e7e18271809c6520, the `,` adds the files on the root level.

R1kaB3rN:
> If I don't misunderstand you, this is already the case with commit [c956ac9](https://github.com/Open-Wine-Components/umu-protonfixes/commit/c956ac9689e59391c6df3380e7e18271809c6520), the `,` adds the files on the root level.

Sorry, my description was sloppy/inaccurate there. Thanks for the correction.

> Well, we could just enforce it on all `*.py` files - without the "static" array.

Yeah and probably define some files to exclude such as `__init__.py` for the gamefixes in [pyproject.toml](https://github.com/Open-Wine-Components/umu-protonfixes/blob/master/pyproject.toml)

Root-Core:
No problem. I moved the linked comment here for better discussion. I also expanded it a bit.

------------

We should enforce this via linter:
- function types
- single quotes
- format strings
- Docstrings for each function (C0116)
- Docstrings indentation on multiple lines
- no blank lines after function Docstrings
- two blank lines between functions
- a reasonable line length (?)

We want to:
- Check if there are unused functions
- Check that GDrive is actually working
- Split large modules like `util.py` into smaller ones
- Enforce linter on all `*.py` files
- Handle the complete linter configuration in the `pyproject.toml` file
- Remove `#pylint: disable=C0103` from gamefixes
- Have as few pylint exceptions as possible

R1kaB3rN:
I think enforcing rules for Docstrings will require additional tools than pylint. 

In addition, there should probably as much tests as possible. In particular for critical files such as `fix.py` and `util.py`

R1kaB3rN:
I have a PR open that's addresses some of these issues.

As for these items:
-     Check that GDrive is actually working
-     Split large modules like util.py into smaller ones

[...]

PR #183, opened 2024-12-19
refactor: prefer pathlib when possible

R1kaB3rN:
Related to https://github.com/Open-Wine-Components/umu-protonfixes/pull/152 and https://github.com/Open-Wine-Components/umu-protonfixes/issues/50.

Refactors to use pathlib when possible or appropriate without directly affecting user-created gamefixes

### DIFF
--- a/fix.py
+++ b/fix.py
@@ -42,42 +42,45 @@ def get_game_id() -> str:
 def get_game_name() -> str:
     """Trys to return the game name from environment variables"""
     pfx = os.environ.get('WINEPREFIX') or protonmain.g_session.env.get('WINEPREFIX')
-    script_dir = os.path.dirname(os.path.abspath(__file__))
+    script_dir = os.path.dirname(os.path.abspath(__file__))  # noqa: PTH100, PTH120
 
     if os.environ.get('UMU_ID'):
-
-        if os.path.isfile(f'{pfx}/game_title'):
-            with open(f'{pfx}/game_title', encoding='utf-8') as file:
+        if os.path.isfile(f'{pfx}/game_title'):  # noqa: PTH113
+            with open(f'{pfx}/game_title', encoding='utf-8') as file:  # noqa: PTH123
                 return file.readline()
 
         umu_id = os.environ['UMU_ID']
         store = os.getenv('STORE', 'none')
-        csv_file_path = os.path.join(script_dir, 'umu-database.csv')
+        csv_file_path = os.path.join(script_dir, 'umu-database.csv')  # noqa: PTH118
 
         try:
-            with open(csv_file_path, newline='', encoding='utf-8') as csvfile:
+            with open(csv_file_path, newline='', encoding='utf-8') as csvfile:  # noqa: PTH123
                 csvreader = csv.reader(csvfile)
                 for row in csvreader:
                     # Check if the row has enough columns and matches both UMU_ID and STORE
                     if len(row) > 3 and row[3] == umu_id and row[1] == store:
                         title = row[0]  # Title is the first entry
-                        with open(os.path.join(script_dir, 'game_title'), 'w', encoding='utf-8') as file:
+                        with open(  # noqa: PTH123
+                            os.path.join(script_dir, 'game_title'),  # noqa: PTH118
+                            'w',
+                            encoding='utf-8',
+                        ) as file:
                             file.write(title)
                         return title
         except FileNotFoundError:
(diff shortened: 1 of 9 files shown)


fix.py
```diff
@@ -42,42 +42,45 @@ def get_game_id() -> str:
...
                     # Check if the row has enough columns and matches both UMU_ID and STORE
                     if len(row) > 3 and row[3] == umu_id and row[1] == store:
                         title = row[0]  # Title is the first entry
-                        with open(os.path.join(script_dir, 'game_title'), 'w', encoding='utf-8') as file:
+                        with open(  # noqa: PTH123
+                            os.path.join(script_dir, 'game_title'),  # noqa: PTH118
+                            'w',
+                            encoding='utf-8',
+                        ) as file:
                             file.write(title)
```
R1kaB3rN, line 68: This might be a bug. If I'm understanding correctly, we're not writing the title in the user's prefix. Instead, the title's is in the protonfixes directory.

fix.py
```diff
@@ -42,42 +42,45 @@ def get_game_id() -> str:
...
                         return title
         except FileNotFoundError:
-            log.warn(f"CSV file not found: {csv_file_path}")
+            log.warn(f'CSV file not found: {csv_file_path}')
```
R1kaB3rN, line 71: Just style changes applied from ruff format.

fix.py
```diff
@@ -129,16 +132,16 @@ def get_module_name(game_id: str, default: bool = False, local: bool = False) ->
...
 
     # Check if local gamefix exists
-    if not os.path.isfile(os.path.join(localpath, module_name + '.py')):
+    if not os.path.isfile(os.path.join(localpath, module_name + '.py')):  # noqa: PTH113, PTH118
         return False
 
     # Ensure local gamefixes are importable as modules via PATH
-    with open(os.path.join(localpath, '__init__.py'), 'a', encoding='utf-8'):
-        sys.path.append(os.path.expanduser('~/.config/protonfixes'))
+    with open(os.path.join(localpath, '__init__.py'), 'a', encoding='utf-8'):  # noqa: PTH118, PTH123
```
R1kaB3rN, line 143: Do we even have to open the file here?

fix.py
```diff
@@ -42,42 +42,45 @@ def get_game_id() -> str:
...
 
         umu_id = os.environ['UMU_ID']
         store = os.getenv('STORE', 'none')
-        csv_file_path = os.path.join(script_dir, 'umu-database.csv')
+        csv_file_path = os.path.join(script_dir, 'umu-database.csv')  # noqa: PTH118
 
         try:
-            with open(csv_file_path, newline='', encoding='utf-8') as csvfile:
+            with open(csv_file_path, newline='', encoding='utf-8') as csvfile:  # noqa: PTH123
                 csvreader = csv.reader(csvfile)
```
R1kaB3rN, line 58: The CSV is a database for us so we handle the CSV by assuming it's very large.

TODO: Refactor to prefer memory mapping the CSV.

fix.py
```diff
@@ -129,16 +132,16 @@ def get_module_name(game_id: str, default: bool = False, local: bool = False) ->
 
 def _run_fix_local(game_id: str, default: bool = False) -> bool:
     """Check if a local gamefix is available first and run it"""
-    localpath = os.path.expanduser('~/.config/protonfixes/localfixes')
+    localpath = os.path.expanduser('~/.config/protonfixes/localfixes')  # noqa: PTH111
```
R1kaB3rN, line 135: Refactoring this line and below as pathlib.Path objects should be safe.

logger.py
```diff
@@ -30,8 +31,9 @@ def log(self, msg: str = '', level: str = 'INFO') -> None:
         fulltext = color + pfx + str(msg) + reset + os.linesep
         sys.stderr.write(fulltext)
         sys.stderr.flush()
-        with open('/tmp/test', 'a', 1, encoding='utf-8') as testfile:
-            testfile.write(logtext)
+
+        with Path('/tmp/test').open(mode='a', encoding='utf-8', buffering=1) as file:
+            file.write(logtext)
```
R1kaB3rN, line 36: We can actually refactor this. But putting that aside why do we need to test if /tmp is writable? We should delete this and prefer adding this as part of our checks.

R1kaB3rN: Never mind. We’re not testing anything here. In those lines,  we’re writing the log message. 

TODO: Refactor to create a secure temporary file.

PR #184, opened 2024-12-20
Automatically install Ubisoft Connect for multiple old Ubisoft games that still ship the uPlay installer

UsernamesAreNotMyThing:
Both games ship with an outdated and non-functional uPlay launcher, that needs to be replaced with Ubisoft Connect, which can be installed through Protontricks.

A lot of reports at https://www.protondb.com/app/33230?device=any and https://www.protondb.com/app/368500?device=any support this (especially the former), and it looks like this fix is still needed.


gamefixes-steam/368500.py
```diff
@@ -0,0 +1,8 @@
+"""Game fix for Assassin's Creed: Syndicate"""
+
+from protonfixes import util
+
+
+def main() -> None:
+	"""Game ships with outdated and not working uPlay launcher."""
+	util.protontricks('ubisoftconnect')
```
R1kaB3rN, line 8: Thanks for this.

Also, following the upstream report for this title, I believe another environment seems to also be needed? Namely, PROTON_DISABLE_NVAPI=1

See https://github.com/ValveSoftware/Proton/issues/2464

UsernamesAreNotMyThing: Going by the second to last comment, it looks like there may already be plans to add that to official Proton.

UsernamesAreNotMyThing:
Looks like Trackmania Turbo also needs this: https://github.com/ValveSoftware/Proton/issues/6637#issuecomment-1852468171

gamefixes-steam/368500.py
R1kaB3rN: We can remove that specific fix once those plans ever come to fruition. For environment variables, there's generally no harm setting it twice. There's only harm when the env var isn't needed anymore and a proper upstream solution has been implemented to resolve the issue. In those cases, we need to pay attention.

gamefixes-steam/375900.py
```diff
@@ -0,0 +1 @@
+33230.py
```
R1kaB3rN, line 1: Create a separate fix for this. I believe we only want to create symbolic links within the same dir if the titles are part of the same series

R1kaB3rN (changes requested):

R1kaB3rN:
For what it's worth, for users playing Trackmania Turbo, I'm referencing the Openplanet issue.

https://github.com/openplanet-nl/issues/issues/609

PR #187, opened 2024-12-27
Workaround around to prevent Sonic Colors Ultimate black screen for some users

UsernamesAreNotMyThing:
I'm not smart enough to know what this is supposed to do, but apparently some people are having trouble with the game getting stuck on a black screen unless the `XDG_DATA_HOME` environment variable is removed.

https://github.com/ValveSoftware/Proton/issues/7538#issuecomment-2562077664

### DIFF
--- a/gamefixes-steam/2055290.py
+++ b/gamefixes-steam/2055290.py
@@ -1,0 +1,8 @@
+"""Sonic Colors: Ultimate"""
+
+from protonfixes import util
+
+
+def main() -> None:
+    """Black screen for some users unless deleting XDG_DATA_HOME variable"""
+    util.del_environment('XDG_DATA_HOME')


R1kaB3rN:
This fix seems off. The issue is graphics related, so that environemnt variable should be irrelevant.

R1kaB3rN:
Furthermore, if we’re to trust ProtonDB, then there appears to be a discrepancy. 

See https://www.protondb.com/app/2055290

So I suspect this is system-specific, not something that is broken for all users.

R1kaB3rN:
Closing as I’m not convinced this game needs to be fixed, and it’s unclear how that env var and the black screen are related.

spiffeeroo:
If you export `$XDG_DATA_HOME` like some people do with ~/.bashrc and ~/.zshrc into your environment, the game experiences those symptoms. An absolutely bone stock config from a fresh install will not experience this issue. The default environment for almost all Linux distos does not set `$XDG_DATA_HOME`. That is introduced by user configuration.

When you add something dumb like `XDG_DATA_HOME="CHILLI-DOGS" %command%` into the launch option in Steam, you will immediately experience the black screen with cursor bug. It will also create this directory at `~/.steam/steam/steamapps/common/SonicColorsUltimate/exec/CHILLI-DOGS/SEGA/SonicColorsUltimate/_yoursteamid_`. The game is supposed to create a folder in the wine prefix at `~/.steam/steam/steamapps/compatdata/2055290/pfx/drive_c/users/steamuser/AppData/Roaming/SEGA/SonicColorsUltimate/Steam/{_yoursteamid_,crashpad}`

Remove that launch option, and the game works normally.

I have Fedora 41 and Arch Linux installed on separate drives. Fedora and Arch Linux were on the same kernel and Mesa version. Once you introduce `$XDG_DATA_HOME` into the environment, the game launch is broken.

PR #191, opened 2024-12-30
FPS limit for Arcana Heart 3

UsernamesAreNotMyThing:
Game runs too fast at high frame rates, and I suspect this is very likely a game bug. The game does have a built-in frame limit, but that only works in fullscreen.

https://github.com/ValveSoftware/Proton/issues/8341#issue-2758322396


R1kaB3rN:
For DXVK-related fixes, please make a fix for it upstream at this file:

https://github.com/doitsujin/dxvk/blob/master/src/util/config/config.cpp

R1kaB3rN:
For a concrete example for limiting FPS for a title, see:

https://github.com/doitsujin/dxvk/blob/1eb0c687a6748d8268dd84f466a1378de4fb3c10/src/util/config/config.cpp#L339

The DXVK devs should be open to accepting it.

UsernamesAreNotMyThing:
Someone's already done this: https://github.com/doitsujin/dxvk/pull/4566

PR #196, opened 2025-01-01
Add directplay for Total Annihilation

kasha695:
When you select Multi and the TCP/IP the game goes back to the Multi menu without directplay.

With directplay it lets you host games and play successfully.

### DIFF
--- a/gamefixes-steam/289030.py
+++ b/gamefixes-steam/289030.py
@@ -1,0 +1,8 @@
+"""Game fix for Total Annihilation"""
+
+from protonfixes import util
+
+
+def main() -> None:
+    """Multiplayer requires directplay for full functionality"""
+    util.protontricks('directplay')


gamefixes-steam/289030.py
```diff
@@ -0,0 +1,8 @@
+"""Game fix for Total Annihilation"""
```
R1kaB3rN, line 1: The app ID is 298030

R1kaB3rN: https://steamdb.info/app/298030/

R1kaB3rN (changes requested):

gamefixes-steam/289030.py
kasha695: Apologies, resolved the issue.

R1kaB3rN (approved):

PR #200, opened 2025-01-08
Refactor: Relative imports / imports without try block

Root-Core:

The unit tests no longer require imports in a try block.
Also, the game fixes imports have been replaced to be relative, as this is necessary for static type checking.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,6 +46,10 @@ jobs:
           ln -s umu-protonfixes protonfixes
           cd protonfixes
           python3 .github/scripts/check_imports.py
+      # We need a known parent package in the unit tests, so we need to change to the parent dir
+      # As "umu-protonfixes" is not a valid package name, we create a symbolic link to "protonfixes"
       - name: Test with unittest
         run: |
-          python3 protonfixes_test.py
+          cd ..
+          ln -s umu-protonfixes protonfixes
+          python3 -m protonfixes.protonfixes_test
--- a/__init__.py
+++ b/__init__.py
@@ -1,18 +1,42 @@
-"""Starts the protonfix module"""
+"""Starts the protonfix module and runs fixes after pre-flight-checks"""
 
 import os
 import sys
+import traceback
 
-RUN_CONDITIONS = [
-    'STEAM_COMPAT_DATA_PATH' in os.environ,
-    'PROTONFIXES_DISABLE' not in os.environ,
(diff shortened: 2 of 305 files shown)


R1kaB3rN (changes requested):
After reslolving the conflict and I'll merge this ASAP

Root-Core:
Done, but it seems that these changes are not compatible with the new CI script.

R1kaB3rN:
Are you sure `from .. import util` is valid? I haven't actually tried running these fixes through Proton.

R1kaB3rN:
I'll have to test locally later.

Root-Core:
Yes, they are just relative imports. I tested them in Proton via. some games in Steam. This also enables Pyright and transforms the code base from a bunch of independent files that somehow references each other to an actual project.

I'm not sure how it even references `protonfixes` in the current implementation.
I guess the python call has the module injected somehow? With the relative imports, there aren't any confusions.

R1kaB3rN:
> Yes, they are just relative imports. I tested them in Proton via. some games in Steam. This also enables Pyright and transforms the code base from a bunch of independent files that somehow references each other to an actual project.
> 
> I'm not sure how it even references `protonfixes` in the current implementation. I guess the python call has the module injected somehow? With the relative imports, there aren't any confusions.

Well,  I can reproduce the error. Change the directory to gamefixes-steam, invoke python, then try to import with `from .. import util`. This will fail.

R1kaB3rN:
I guess another way is to use umu-launcher and run all the fixes through it which I have been avoiding. Because of course, it'll take a while though even when running it concurrently.

Root-Core:
> Well, I can reproduce the error. Change the directory to gamefixes-steam, invoke python, then try to import with `from .. import util`. This will fail.

This is expected behaviour in Python, but imports can get ugly pretty fast there anyways.
That's why we launch the unit test from outside of the module.. or we need to fixup the imports via an try block.

> I guess another way is to use umu-launcher and run all the fixes through it which I have been avoiding. Because of course, it'll take a while though even when running it concurrently.

I don't think this is necessary, as pyright is a well known and maintained solution.
I'm not a huge fan of spinning an home-brewed solution for something that's already solved.

Nonetheless, you would only need to run it once over the full set of fixes. The github workflow should give you a list of changed files, as far as I know. This could be used for incremental checks.

Root-Core:
Closing in favor of #271
We could refactor the fixes to use relative imports later.. or just omit that step altogether, as #205 should now resolve the module via it's config.

PR #201, opened 2025-01-09
Refactor: Fixed and enforced types / logger class rewrite + CI: enabled static type checking

Root-Core:

This PR fixes issues that were found in the static typing analysis and other bad practices / issues with types.

I included the rather trivial rewrite of the logger class.
It uses constants quite a lot and benefits greatly from the shift towards enums.

### DIFF
--- a/.github/scripts/check_verbs.py
+++ b/.github/scripts/check_verbs.py
@@ -4,16 +4,16 @@ import os
 import re
 import subprocess
 
-from glob import iglob
 from pathlib import Path
 from tempfile import mkdtemp
+from collections.abc import Generator
 
 # 'gui' is a virtual verb for opening the Winetricks GUI
 # 'vd=1280x720' is a setting for the virtual desktop and valid
 whitelist_verbs = {'gui', 'vd=1280x720'}
 
 
-def extract_verbs_from_glob(path_glob: iglob) -> set[str]:
+def extract_verbs_from_glob(path_glob: Generator[Path, None, None]) -> set[str]:
     """Simply strip the extension from all found files."""
     return {file.stem for file in path_glob}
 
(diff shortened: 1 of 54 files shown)


GloriousEggroll:
I like the logging changes, I do not like the winedll_override changes. 

The reason the winedll_override is the way it currently is is because it's akin to how it's done in winetricks and users who are familiar with wine are familiar with this format, particularly with setting b/builtin,n,native.

Root-Core:
While I understand your point, I have found several mistakes made by having this parameter type as a string. An enum prevents errors altogether and is a clean way to protect yourself when implementing a change.

Without prior knowledge, a user would still have to copy the line or at least read the function definition. So I don't think this will be a problem. Also, the function and enum are now well documented. It would be better if IDEs could resolve the modules and show intellisense hints, but we would have to change _all_ fixes to use relative imports of `util`, just like #200 did. (Which might still be a good idea, idk.)

GloriousEggroll:
> While I understand your point, I have found several mistakes made by having this parameter type as a string. An enum prevents errors altogether and is a clean way to protect yourself when implementing a change.
> 
> Without prior knowledge, a user would still have to copy the line or at least read the function definition. So I don't think this will be a problem. Also, the function and enum are now well documented. It would be better if IDEs could resolve the modules and show intellisense hints, but we would have to change _all_ fixes to use relative imports of `util`, just like #200 did. (Which might still be a good idea, idk.)

what about instances where both native and builtin are being set and we just are rearranging the order? How would the enums work there?

-edit-

nvm i see NATIVE_BUILTIN

Can we change the prefix to WINEDLLOverride instead of PE...? I think thats more familiar to people.

Root-Core:
I rebased again and changed the name to `OverrideOrder`.
If you can come up with a better name, I can still rename it. :)

PR #203, opened 2025-01-09
Refactor:  prefer pathlib and general cleanup

Root-Core:

There might some conflicts with the other PRs, but they should be trivial to resolve.

This PR prefers pathlib over joined strings and uses it's API to prevent bugs and code smells.
Some common functions are also moved out of individual fixes and now available in `utils.py`.
Some things are reimplemented and some are cleaned up.


GloriousEggroll:
I think we need to separate the pathlib refactor from the other 'general cleanup' tasks and submit the general cleanup tasks as separate MRs. I don't want to submit this is as one big jumbled mess and then have to sort through it if something is broken

Root-Core:
Sure, I'm a bit busy atm, so it might take some time. This is more or less just refactoring, so not too important.
Fixing and preventing type errors and bugs should be a priority (#201).

R1kaB3rN:
I'll have to agree with GE on this one.

At it's current state, this merge request isn't acceptable due how wide spread the changes are and the lack of tests. In general, I'm **not** comfortable with any "cleanup" changes to our user-created gamefixes unless there are clearly malicious/bad practices in them (e.g., forgetting to close file descriptors, dubiously reading irrelevant files in $HOME, etc.). Or alternatively were auto-fixed from a sane, automated tool which is not the case here. But more on the topic of this merge request, we do not gain anything signficiantly beneficial when transitioning to pathlib besides the risk of syntax/logical errors from this refactor as we don't test most of the functionality directly to ensure the previous behavior had been preserved besides in fix.py and that no new bugs were introduced. That's the issue here even if they are "trivial" to resolve, so I'm compelled to assume the worse from these changes.

Maybe if this were a personal project or a project with ample resources that has a very small user base who doesn't mind some breakage it would be ok make changes like this. But not in this project when Lutris, HGL, and other clients are using this module indirectly and are expecting things to just work.

Between accepting the current code smells and this merge request, I would rather keep things as-is unless a separate pull request was made with tests, and was both focused and narrow in scope. Of course, you're not obligated to do this and I understand it would be a big request to make just for a FOSS project :ok_hand:

PR #204, opened 2025-01-09
ci: validate import statements in modules

R1kaB3rN:


Root-Core:
Pyright already validates imports. What exactly are you trying to achieve?

R1kaB3rN:
> Pyright already validates imports. What exactly are you trying to achieve?

I'm looking to validate imports but merging the import changes. Yeah, pyright would be better. Would you just rename the project directory then run it? I get reportMissingImports errors doing it that way?

Root-Core:
#152 implements pyright in the workflow and it's working without any problem.. as long, as my changes are applied.

Most "import"antly #200, as this fixes the issues that pyright has with the the imports.

R1kaB3rN:
Yeah, I want to merge 200 after having this check is in the CI. Because in 200, pyright isn't in yet.

Root-Core:
We could merge pyright and and limit it's scope / files to something that works for us. If it's only about the imports, we could just use the rules for the imports I guess.

R1kaB3rN:
> Meanwhile https://github.com/Open-Wine-Components/umu-protonfixes/pull/152 is the prove, that https://github.com/Open-Wine-Components/umu-protonfixes/pull/200 is fine in regards of it's imports. The CI already verified them, as Pyright is active there and https://github.com/Open-Wine-Components/umu-protonfixes/pull/200 is just a cherry-picked subset.

Yeah, that logic makes sense...

R1kaB3rN:
> EDIT: I just tested it and can confirm, that setting `typeCheckingMode = "off"` in the `pyproject.toml` checks the imports and works with the current code base. `ijson` and `steam` need to be installed though.

Mind referencing the docs for that? I can't find info about that value.

R1kaB3rN:
Well, in any case, I want to get this in for now. We can remove it later for pyright, which is faster.

Root-Core:
Sure.. I created PR #205, just have a look. It currently annotates the imports and flags it as a warning, but we could let the CI fail. 

About the documentation:
https://github.com/microsoft/pyright/blob/main/docs/configuration.md#type-check-diagnostics-settings

PR #205, opened 2025-01-09
CI: Annotations for shellcheck and ruff + enable pyright (without static type checking)

Root-Core:

As discussed in #204 Pyright is able to check the imports, [without checking types](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#type-check-diagnostics-settings). I'm not sure what else is tested, but imports are.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,11 +33,13 @@ jobs:
 
       # FIXME: problem matcher is currently disabled upstream, using a fork for the moment
       # https://github.com/ludeeus/action-shellcheck/pull/103
+      # FIXME: symlinks don't work upstream
+      # https://github.com/ludeeus/action-shellcheck/pull/104
       - name: Run ShellCheck
-        uses: Root-Core/action-shellcheck@master
+        uses: Root-Core/action-shellcheck@fork
         with:
-          format: gcc # gcc is used for the problem matcher
-          additional_files: winetricks # by default, only files with according extensions are checked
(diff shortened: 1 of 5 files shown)


R1kaB3rN:
Why doesn't pyright error on resolution?

R1kaB3rN:
I expected the warning:

`Import "protonfixes" could not be resolved (reportMissingImports)`

to have the workflow fail, which doesn't. I guess it's just a matter of configuring pyright?

Root-Core:
Github checks against unchanged files I guess. Therefore these are only warnings.. or the warning level is not high enough / pyright doesn't fail.

R1kaB3rN:
> Github checks against unchanged files I guess. Therefore these are only warnings.. or the warning level is not high enough / pyright doesn't fail.
> 
> > 0 errors, 392 warnings, 0 informations
> > ··• echo $?
> > 0
> 
> Yeah.. they aren't errors. I think we could configure pyright to handle them as errors.

Yeah, we can look into that later.

Root-Core:
Using the symbolic link in combination with `extraPaths[]` works.
I still think the relative imports are a better solution, but this does the trick for now and we could skip most of #200.

Or did you test the changes in #200 and concluded that they also work and are indeed a "better" solution? :)

GloriousEggroll:
@Root-Core do you have access or do you want access to our umu dev discord channel? you've been a contributor for a long time, i think it would be beneficial. if you can add me on Discord and tell me what your tag is I can send an invite

PR #231, opened 2025-02-18
Handle PermissionError in check_internet

PapayaJackal:
Updated the `check_internet` function to account for `PermissionError` exception that may occur when the firewall is set to block the connection to 1.1.1.1.


R1kaB3rN:
Sure, but do you mind sharing the traceback and possibly simple steps to reproduce the exception?

PR #243, opened 2025-02-27
Liminal Border Part III

knosuhor:
Only needed for videos to play. Could be used for Part I and II.

### DIFF
--- a/gamefixes-steam/3094040.py
+++ b/gamefixes-steam/3094040.py
@@ -1,0 +1,7 @@
+from protonfixes import util
+
+
+def main() -> None:
+    util.protontricks('wmp9')
+    util.protontricks('directshow')
+    util.disable_protonaudioconverter()


gamefixes-steam/3094040.py
```diff
@@ -0,0 +1,7 @@
+from protonfixes import util
+
+
+def main() -> None:
+    util.protontricks('wmp9')
```
R1kaB3rN, line 5: Just the `wmp11` verb should suffice.

gamefixes-steam/3094040.py
```diff
@@ -0,0 +1,7 @@
...
+    util.protontricks('directshow')
+    util.disable_protonaudioconverter()
```
R1kaB3rN, line 7: Remove lines 6 and 7. Should be unnecessary.

R1kaB3rN (changes requested):

R1kaB3rN:
> Only needed for videos to play. Could be used for Part I and II.

Sure. We can add fixes for the other games as well.

R1kaB3rN:
Will merge ASAP once you make those changes.

gamefixes-steam/3094040.py
knosuhor: @R1kaB3rN Tried that before, it crashes with wmp11.

gamefixes-steam/3094040.py
knosuhor: Tried that, it  crashed on video playback with wmp11 and froze with just wmp9.

R1kaB3rN:
https://github.com/ValveSoftware/Proton/issues/8485

PR #248, opened 2025-03-02
Add fix for Deus Ex: Invisible War extremely loud audio glitches

eskay993:
This fixes the extremely loud audio glitches that happen at various points in the game, most notably during the hangar section and when using Spy Drones.


R1kaB3rN:
https://github.com/ValveSoftware/Proton/issues/3627

PR #256, opened 2025-03-15
Add DLL override to fix modding in Elin

UsernamesAreNotMyThing:
https://github.com/ValveSoftware/Proton/issues/8522#issue-2919285714


R1kaB3rN:
The game runs fine without this change, but this would be a nice to have.

cc @Tiagoquix for thoughts, given your involvement in https://github.com/ValveSoftware/Proton/issues/8308#issuecomment-2549603517. Just curious to know if you're opinion has changed regarding the winhttp.dll.

Issue #259, opened 2025-03-20
Can not find game in *umu-database.csv*


SilPan:
**Describe the bug**

[`fix.py:get_game_name`](https://github.com/Open-Wine-Components/umu-protonfixes/blob/f18ac154d12f1e1af808e0246c2213790bbec929/fix.py#L42C5-L42C18) fails to find the game, even though the environment variable `UMU_ID` is set.

I already had a PR ready to switch from `csv.reader` to `csv.DictReader`, because the logic had me confused when it turned out that this was not the issue..
For some reason if the other environment variable `STORE` is not set, **then** the title won't be found.

What is supposed to happen is that if STORE is not set it should default to `none`.
Which is exactly what the line `store = os.getenv('STORE', 'none')` does.

For some reason `store` stays empty (I am really dumbfounded why) and the game does not get found.

**To Reproduce**
Have `UMU_ID` set and `STORE` not.

**Expected behavior**
Find games also if `STORE` is not set.

SilPan:
[This](https://github.com/Open-Wine-Components/umu-protonfixes/pull/260) is the mentioned PR.

R1kaB3rN:
How were you testing this exactly? To note, this is expected behavior if STORE is set in the environment with an empty string. You confirm this yourself after opening a terminal and starting an interactive Python shell.

R1kaB3rN:
Concretely, here's the relevant line:

https://github.com/Open-Wine-Components/umu-protonfixes/blob/f18ac154d12f1e1af808e0246c2213790bbec929/fix.py#L53

The logical assigment should be:

```
store = os.getenv('STORE') or 'none'
```

So I do think you're right that this is a bug. If make this change locally, I do believe it should solve the issue for you

R1kaB3rN:
> First the change does not do anything because store = os.getenv('STORE', 'none')  and store = os.getenv('STORE') or 'none' are syntactically equivalent, right?

They’re neither syntactically nor logically equivalent if STORE is an empty string, which umu-launcher can do. You can prove this in an interactive Python shell.

PR #260, opened 2025-03-20
switch from csv.csvreader to csv.DictReader

SilPan:
This is the PR mentioned in https://github.com/Open-Wine-Components/umu-protonfixes/issues/259.

It does not really change anything.. I just got very confused when debugging something there.
This version does *not* have magic numbers for csv row ids.

### DIFF
--- a/fix.py
+++ b/fix.py
@@ -55,18 +55,16 @@ def get_game_name() -> str:
 
         try:
             with open(csv_file_path, newline='', encoding='utf-8') as csvfile:
-                csvreader = csv.reader(csvfile)
+                csvreader = csv.DictReader(csvfile)  # Reads CSV as a dictionary per row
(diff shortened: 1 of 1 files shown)


R1kaB3rN:
This change to the implementation appears to introduce a regression.

Putting this change aside, a memory-mapped solution would be preferred since we're reading from a database that will only become larger. If you know how to do that and don't mind, that'd be great. Otherwise, it's fine :ok_hand:

SilPan:
Yes sorry, I just had already written the PR before I realised this is not the issue.
Throw it away.

I'd be interestedbrhough if someone can reproduce my issue or if it is only me.

I'll have a look thougj on how to parse the "database" more efficiently though. I had a quick look before I wrote this, but it was not the focus.

Thank you for maintaining this!

PR #262, opened 2025-03-24
only say 'we did not find the game' if title is still empty after searching the csv

SilPan:
This is the actual fix for the behavior I observed in https://github.com/Open-Wine-Components/umu-protonfixes/issues/259

### DIFF
--- a/fix.py
+++ b/fix.py
@@ -61,7 +61,8 @@ def get_game_title(pfx: str, database: str) -> str:
     except Exception as ex:
         log.debug(f'Error reading CSV file: {ex}')
 
-    log.warn('Game title not found in CSV')
+    if title=='UNKNOWN':
+        log.warn('Game title not found in CSV')
 
     return title
 


fix.py
```diff
@@ -61,7 +61,8 @@ def get_game_title(pfx: str, database: str) -> str:
     except Exception as ex:
         log.debug(f'Error reading CSV file: {ex}')
 
-    log.warn('Game title not found in CSV')
+    if title=='UNKNOWN':
```
R1kaB3rN, line 64: Right, the warning shouldn’t be logged if a title was found. 

Would prefer a change where the function returns the title immediately instead:

```
return row[0]
```

R1kaB3rN (changes requested):

R1kaB3rN (approved):
No worries, I'll make the change myself :ok_hand:. Otherwise, it's good.

PR #265, opened 2025-03-27
Wuthering Waves (3513350) fixes

cybik:
* Rename `winedll_override` to `winepe_override` to make it reusable with new utility methods `winedll_override` and `wineexe_override`
* Add Wuthering Waves fixes (provisional, based on current game runtime), required workarounds notwithstanding

### DIFF
--- a/util.py
+++ b/util.py
@@ -410,14 +410,22 @@ def get_game_install_path() -> str:
     return install_path
 
 
-def winedll_override(dll: str, dtype: Literal['n', 'b', 'n,b', 'b,n', '']) -> None:
-    """Add WINE dll override"""
-    log.info(f'Overriding {dll}.dll = {dtype}')
-    setting = f'{dll}={dtype}'
+def winepe_override(target: str, filetype: str, dtype: Literal['n', 'b', 'n,b', 'b,n', '']) -> None:
+    """Add WINE file override"""
+    log.info(f'Overriding {target}.{filetype} = {dtype}')
+    target_file = target if filetype == "dll" else f'{target}.{filetype}'
+    setting = f'{target_file}={dtype}'
     protonmain.append_to_env_str(
         protonmain.g_session.env, 'WINEDLLOVERRIDES', setting, ';'
     )
 
+def winedll_override(dll: str, dtype: Literal['n', 'b', 'n,b', 'b,n', '']) -> None:
+    """Add WINE dll override"""
+    winefile_override(target=dll, filetype="dll", dtype=dtype)
+
+def wineexe_override(exe: str, dtype: Literal['n', 'b', 'n,b', 'b,n', '']) -> None:
+    """Add WINE executable override"""
+    winefile_override(target=exe, filetype="exe", dtype=dtype)
 
 def patch_libcuda() -> bool:
     """Patches libcuda to work around games that crash when initializing libcuda and are using DLSS.
--- a/util.py
+++ b/util.py
@@ -421,11 +421,11 @@ def winepe_override(target: str, filetype: str, dtype: Literal['n', 'b', 'n,b', 
 
 def winedll_override(dll: str, dtype: Literal['n', 'b', 'n,b', 'b,n', '']) -> None:
     """Add WINE dll override"""
-    winefile_override(target=dll, filetype="dll", dtype=dtype)
+    winepe_override(target=dll, filetype="dll", dtype=dtype)
 
 def wineexe_override(exe: str, dtype: Literal['n', 'b', 'n,b', 'b,n', '']) -> None:
     """Add WINE executable override"""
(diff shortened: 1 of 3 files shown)


R1kaB3rN:
I don't think it makes sense to have a fix for this title when it isn't released yet on Steam. There isn't even a public demo available either. 

Instead, we can create the fix in `umu-protonfixes` instead for the standalone installer that one can get from the official website. Presumably, that's the one you tested against, correct?

util.py
```diff
@@ -410,14 +410,22 @@ def get_game_install_path() -> str:
     return install_path
 
 
-def winedll_override(dll: str, dtype: Literal['n', 'b', 'n,b', 'b,n', '']) -> None:
-    """Add WINE dll override"""
-    log.info(f'Overriding {dll}.dll = {dtype}')
-    setting = f'{dll}={dtype}'
+def winepe_override(target: str, filetype: str, dtype: Literal['n', 'b', 'n,b', 'b,n', '']) -> None:
```
R1kaB3rN, line 413: Is this intentional? You created a new function but it appears you aren't referencing it anywhere.

R1kaB3rN (changes requested):

util.py
```diff
@@ -410,14 +410,22 @@ def get_game_install_path() -> str:
...
+    log.info(f'Overriding {target}.{filetype} = {dtype}')
+    target_file = target if filetype == "dll" else f'{target}.{filetype}'
+    setting = f'{target_file}={dtype}'
     protonmain.append_to_env_str(
         protonmain.g_session.env, 'WINEDLLOVERRIDES', setting, ';'
     )
 
+def winedll_override(dll: str, dtype: Literal['n', 'b', 'n,b', 'b,n', '']) -> None:
+    """Add WINE dll override"""
+    winefile_override(target=dll, filetype="dll", dtype=dtype)
```
R1kaB3rN, line 424: Are we missing commits? This function isn't referenced anywhere within this module's global scope

R1kaB3rN (changes requested):

util.py
cybik: Unintentional, I had renamed the method locally and had forgotten to commit the rename properly. It was the winefile method referred below.

util.py
cybik: See above.

cybik:
> Since this is media playback-related, by any chance, have you already tried disabling Proton's media converter by calling `util.disable_protonmediaconverter()`?

I have, and it didn't do anything to fix the issue. The ***only*** thing that worked is throwing mfplat out unceremoniously.

> I don't think it makes sense to have a fix for this title when it isn't released yet on Steam. There isn't even a public demo available either.

Technically fair, but I think being prepared for it is fine since [we know its ID](https://store.steampowered.com/app/3513350/Wuthering_Waves/) and it'll presumably share the same issues.

> Instead, we can create the fix in umu-protonfixes instead for the standalone installer that one can get from the official website. Presumably, that's the one you tested against, correct?

Pretty much, yeah. I have local launch scripts that set up the right overrides and values (including the Steam game ID, now that we have it) to make it work through Proton (with the "extra sauce" mentioned in [the ticket](https://github.com/ValveSoftware/Proton/issues/8569)).

cybik:
Added lint fixes

R1kaB3rN:
> Technically fair, but I think being prepared for it is fine since [we know its ID](https://store.steampowered.com/app/3513350/Wuthering_Waves/) and it'll presumably share the same issues.

I don’t think we should presume when it’s not released for that storefront yet. You made a report upstream so it’s equally likable that they see it and try to address it before the release that’s in 4 weeks, right? 

Moreover, you tested against the official launcher binaries. So it only makes sense the fix is in `gamefixes-umu` for now and we add it later.

util.py
R1kaB3rN, line 413: Perhaps we should prefix the function with an “_” to indicate it’s internal? E.g.,

```
def _winepe_override
```

Unless there’s a case where a user would invoke this directly? If so, we would need to refactor the logic to handle all file type cases.

R1kaB3rN (changes requested):

util.py
cybik: It's a utility function so it *shouldn't* be exposed.

GloriousEggroll:
Going to agree here to move the fix to gamefixes-umu (or whatever platforms it is currently on) then we can add it to steam later and make our usual symlinks when it's actually in steam

cybik:
All right then: https://github.com/Open-Wine-Components/umu-database/pull/85

GloriousEggroll:
bleh it not being on steam yet makes naming a mess, i just kept the steam id for now. closing this pr because I merged + fixed conflicts via cherry-pick instead.

cybik:
The fix will become moot with Proton 10!

PR #273, opened 2025-03-31
fix: import config modules relatively

R1kaB3rN:
Attempts to fix the import error introduced by 889a5b1c45043d94d37f08fbad1e2cb773a9c692


fix.py
```diff
@@ -13,9 +13,9 @@
     from .checks import run_checks
     from .logger import log
 except ImportError:
-    from config import config
-    from checks import run_checks
-    from logger import log
```
R1kaB3rN, line 18: Commit https://github.com/Open-Wine-Components/umu-protonfixes/commit/889a5b1c45043d94d37f08fbad1e2cb773a9c692 actually causes the interpreter to raise a ModuleNotFoundError.

Refactoring the entire try/except block to only import modules relatively would suffice instead

R1kaB3rN: Wil refactor in a future commit

PR #277, opened 2025-04-03
Refactor: some "trivial" changes

Root-Core:
I've extracted some code from #203 that doesn't have much to do with pathlib stuff.
Most of the changes are "less than trivial" and the gamefixes are more or less unchanged.

Changes to gamefixes:
- 508980
    - Changed string concatenation to use of pathlib
- 2475980 (Gobliiins 5)
    - reimplemented, tested and extended to handle the demo version (2505910)
- 294700 / 497360
    - moved the dos-device creation to `util.py`
    - no functional change
- 22330
    - moved `class Redirect` to `util.py` as `class ReplaceType`
    - type will be reused in later patches
    - no functional change
- 497360
    - Reduced nesting
    - no functional change
- 39500 / 65600
    - Unified
    - Removed unnecessary path join
    - no functional change


Root-Core:
Some quick debugging:

The CI error does not seem to be related to the PR.
We call `get_valid_appids()` several times and at one point the server returns this:

> {'meta_data_only': True, 'response_pending': False, 'http_min_size': 1024, 'http_host': 'clientconfig.akamai.steamstatic.com'}

Which obviously doesn't contain the key `apps`. So we should check if `apps` exists or handle the error somehow.
More research needed.

PR #278, opened 2025-04-04
`SteamDeck=1` for The Last of Us Part 2 Remastered

UsernamesAreNotMyThing:
Looking at https://github.com/ValveSoftware/Proton/issues/8585#issuecomment-2776568529, it looks like this game is seeing a similar issue to God of War Ragnarok, where it fails to start due to an error with PSSDK, despite PSN being optional for this game. This workaround, however, does not seem to work for everybody, and there's currently no explanation for why this is.

### DIFF
--- a/gamefixes-steam/2531310.py
+++ b/gamefixes-steam/2531310.py
@@ -1,0 +1,8 @@
+"""The Last of Us Part 2 Remastered"""
+
+from protonfixes import util
+
+
+def main() -> None:
+    """Fixes the game not starting for some people, but it doesn't work for everyone though. Cause has yet to be determined."""
+    util.set_environment('SteamDeck', '1')


R1kaB3rN:
This game just released and if we're to take that one comment seriously, I think it would be best to wait a bit for a definitive solution rather than a fix that sometimes works.

HunterCZ122:
Works for me out of the box without SteamDeck=1 on AMD GPU.

R1kaB3rN:
Closing as I’m not convinced this is an issue.

The only real issue are the graphics-related ones which should have been addressed in https://github.com/jp7677/dxvk-nvapi/commit/17342694868e661589e3f10012fd56a0936bc5db. Therefore, GE-Proton will need to be updated to at least https://github.com/ValveSoftware/Proton/commit/c44d0d8239ac90e11bb2c1201a1e798f69cc9b99.

Issue #283, opened 2025-04-10
Lord Of The Rings Online mouse fix


mdmatthias:
LOTRO has a bug in newer proton versions where the mouse is not hidden when you rotate the camera. It is fixed with the script mentioned here: https://www.protondb.com/app/212500#riW7--xEaK
In Heroic Launcher I added this script as pre launch script (wrapped in a .sh script to run in background). Is it possible to integrate this in umu-protonfixes? Or some better way to fix it?

R1kaB3rN:
For this title, rather than that program, it should be possible to reimplement the fix in Python using python-xlib or similar. Another way would be to use gamescope as suggested by the upstream issue tracker or make a wine patch just for this title.

mdmatthias:
> For this title, rather than that program, it should be possible to reimplement the fix in Python using python-xlib or similar. Another way would be to use gamescope as suggested by the upstream issue tracker or make a wine patch just for this title in GE-Proton/UMU-Proton

Ok I will try to get it working in Python.

mdmatthias:
Got it working in python:
`import struct
import sys
import time
from Xlib.X import LASTEvent
from Xlib.display import Display
from Xlib.ext import xinput


def is_window_focused(dpy, window):
    """Returns True if the given window is currently focused (active)."""
    root = dpy.screen().root
    net_active_window = dpy.get_atom('_NET_ACTIVE_WINDOW')
    prop = root.get_full_property(net_active_window, 0)
    if not prop:
        return False
    return prop.value[0] == window.id


def get_window_name(dpy, win):
    """Retrieve the window name using WM_NAME or _NET_WM_NAME."""
    name = win.get_wm_name()
    if name:
        return name
    net_name = win.get_full_property(dpy.get_atom('_NET_WM_NAME'), 0)
    if net_name:
        return net_name.value.decode('utf-8')
    return None


def find_window_by_title(dpy, title, win=None):
    """Recursively find a window with a title containing the given string."""
    if win is None:
        win = dpy.screen().root
    name = get_window_name(dpy, win)
    if name and title in name:
        return win
    for child in win.query_tree().children:
        found = find_window_by_title(dpy, title, child)
        if found:
            return found
    return None


def get_game_window(dpy, title):
    game_window = None
    while game_window is None:
        game_window = find_window_by_title(dpy, title)
        if game_window is None:
            time.sleep(1)
    return game_window


def main():
    title = "The Lord of the Rings Online"
[...]

PR #288, opened 2025-04-18
Mouse fix The Lord of the Rings Online, 212500, umu-212500

mdmatthias:
@R1kaB3rN this should be fully working for the mouse fix, tested in latest proton-ge in heroic launcher.
I fixed it with a subprocess. With a daemon thread it was not possible, because a daemon thread also gets killed when the main script ends. The subprocess keeps running when the main script ends, and will exit by itself when the game window gets closed.

closes https://github.com/Open-Wine-Components/umu-protonfixes/issues/283

### DIFF
--- a/gamefixes-steam/212500.py
+++ b/gamefixes-steam/212500.py
@@ -1,7 +1,7 @@
 """Game fix The Lord of the Rings Online"""
-
-#
-from protonfixes import util
+import os
+import subprocess
+import sys
 
 import struct
 import time
@@ -35,7 +35,7 @@ def get_window_name(dpy: Display, win: Window) -> Union[GetProperty, None]:
 
 
 def find_window_by_title(
-    dpy: Display, title: str, win: Union[Display, None] = None
+        dpy: Display, title: str, win: Union[Display, None] = None
 ) -> Union[Window, None]:
     """Recursively find a window with a title containing the given string."""
     if win is None:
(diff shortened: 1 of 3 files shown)


gamefixes-steam/212500.py
```diff
@@ -1,11 +1,166 @@
...
+            retries += 1
+            if retries >= 30:
+                print("Game window not found for 30 seconds, closing mouse fix.")
+                # close display before exit
+                dpy.close()
+                exit()
+    return game_window
+
+
+def mouse_fix(title: str) -> None:
```
R1kaB3rN, line 70: To be clear, you were unsuccessful in running this function in the background? For example:

```
import threading
threading.Thread(target=mouse_fix, title='The Lord of the Rings Online', daemon=True).start()
```

R1kaB3rN (changes requested):

gamefixes-steam/212500.py
R1kaB3rN: >  With a daemon thread it was not possible, because a daemon thread also gets killed when the main script ends.

Please substantiate your claim by pasting a traceback of the daemon thread being killed after running this game through Heroic with umu enabled. This should work. For example:

umu-foo.py:
```
import threading
import time

def foo() -> None:
    while True:
        print('foo')
        time.sleep(1)
        
def main() -> None:
    threading.Thread(target=foo, daemon=True).start()
```

```
> GAMEID=umu-foo PROTONPATH=$HOME/.local/share/umu/compatibilitytools/GE-Latest UMU_RUNTIME_UPDATE=0 umu-run winecfg
```

I would expect daemon thread to still be running either until the interpreter exits, is explicitly killed by an external process, or when the game exits. 

Alternatively to using threads, another approach would be to write the mouse fix functionality to a temporary file and create a subprocess where the interpreter runs it.

R1kaB3rN:
This is good enough. We can ignore the CI errors for now. python-xlib is now being built to support this fix, and I'll update it myself to ensure that it's being imported.

R1kaB3rN:
Linking upstream report: https://github.com/ValveSoftware/Proton/issues/1816

mdmatthias:
@R1kaB3rN Looks like this is no longer needed in ge-proton 10. In ge-proton 10 it works as it should without enabling the fix.
Also when enabling error logs ( by removing stderr=subprocess.DEVNULL in subprocess ), I see this:
```

  File "/tmp/tmpsmyp16e3.py", line 154, in <module>

    sys.exit(_mouse_fix_subprocess())

  File "/tmp/tmpsmyp16e3.py", line 11, in _mouse_fix_subprocess

    from Xlib.display import Display

  File "/home/matthias/.config/heroic/tools/proton/GE-Proton10-1/protonfixes/_vendor/Xlib/display.py", line 26, in <module>

    from six import create_unbound_method

ModuleNotFoundError: No module named 'six'
```
So my fix doesnt even work anymore in ge-proton 10. It did work when I pulled the latest umu proton fixes into ge-proton-27. But no longer need to find out why since it works without it. But better remove it since my fix is worse than the current working one, my fix was always hiding the mouse when you hold mousebuttons. But the current working fix in proton 10 is just as the game intended and really only hiding it when rotating camera.

R1kaB3rN:
Great. Nice to know that’s it’s working in 10. 

Yeah, the reason why the fix doesn’t work anymore is due to last minute changes that were merged related to the build for umu-protonfixes. If you’re interested in the details, just look at the Makefile commit history.

PR #289, opened 2025-04-18
fix check_conditions in __init__ when there is only 1 sys.argv

mdmatthias:
the original check can trigger this error:
'waitforexitandrun' in sys.argv[1]
IndexError: list index out of range

to select [1] you need atleas a len of 2

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -16,7 +16,7 @@ def check_conditions() -> bool:
 
     """
     return (
-        len(sys.argv) >= 1
+        len(sys.argv) >= 2
         and 'STEAM_COMPAT_DATA_PATH' in os.environ
         and 'PROTONFIXES_DISABLE' not in os.environ
         and 'waitforexitandrun' in sys.argv[1]


__init__.py
```diff
@@ -16,7 +16,7 @@ def check_conditions() -> bool:
 
     """
     return (
-        len(sys.argv) >= 1
```
R1kaB3rN, line 19: I'm interested in more context here. Do you mind listing steps to reproduce this error?

R1kaB3rN (changes requested):

__init__.py
mdmatthias: I noticed it while doing this fix for lotro https://github.com/Open-Wine-Components/umu-protonfixes/pull/288
It runs like this in a subprocess "python 212500.py" so then sys.argsv looks like this:
["212500.py"] -> so that means sys.argsv[0] = "212500.py" and sys.argsv[1] will not be there -> IndexError

To reproduce it you can just create a python file with only this import:
```
from protonfixes import util
```
then run that python file like this to trigger the error: 
```
STEAM_COMPAT_DATA_PATH=somevalue python test_argv_bug.py
```
result:
```
ProtonFixes[21899] WARN: [CONFIG]: Parent directory "/home/matthias/.config/protonfixes" does not exist. Abort.
  File "/home/matthias/PycharmProjects/test_argv_bug.py", line 1, in <module>
    from protonfixes import util
  File "/home/matthias/PycharmProjects/protonfixes/__init__.py", line 38, in <module>
    elif not check_conditions():
             ~~~~~~~~~~~~~~~~^^
  File "/home/matthias/PycharmProjects/protonfixes/__init__.py", line 22, in check_conditions
    and 'waitforexitandrun' in sys.argv[1]
                               ~~~~~~~~^^^
IndexError: list index out of range
```

```
  return (
        len(sys.argv) >= 1 # this will be a length of 1 -> True
        and 'STEAM_COMPAT_DATA_PATH' in os.environ # will be True
        and 'PROTONFIXES_DISABLE' not in os.environ # will be True
        and 'waitforexitandrun' in sys.argv[1] # since length is only 1, sys.argv[1] will be out of range
    )
```
So for this check to work correctly the length of argv should be atleast 2. Games probably always start with more than 1 argument, so thats why it never triggered an error before. To workaround this in my lotro fix i moved the import into the main() function, which triggers when game starts, so then we dont have that error during the subprocess, since the subprocess only uses the main_background() function.

R1kaB3rN: No workaround is needed here. Under normal circumstances, an IndexError should not be raised and the reason that it occurs here is due to the your incorrect usage:

```
STEAM_COMPAT_DATA_PATH=somevalue python test_argv_bug.py
```

You don't run Proton games or fixes like this from the command line, even for testing purposes. You should do it through a client that supports running Proton correctly or by following the [documentation](https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/-/blob/main/docs/slr-for-game-developers.md#running-a-game-under-proton-in-the-steam-linux-runtime-environment).

When running games through Proton correctly,  a verb, usually `waitforexitandrun`,  is apart of the command line arguments unless configured not to. After the namespace changes, the arguments would be something at the proton script:

```
['/home/foo/.local/share/umu/compatibilitytools/GE-Latest/proton', 'waitforexitandrun', 'winecfg']
```
Where sys.argv[1] is a verb, and where sys.argv contains at least 2 non-empty values. There's never a case where there's less than 3 arguments  or when a verb is not set.

PR #294, opened 2025-04-21
update util.patch_libcuda

mactan-sc:
- Star Citizen's easy anti-cheat clears values from LD_PRELOAD such that the patched libcuda is not used. Instead use LD_LIBRARY_PATH which is not presently overridden. 
- The game checks for the existence of these dlls when trying to initialize ngx. To satisfy this validation an existing dll is copied to the three names
  -  cryptbase.dll
  - devobj.dll
  - drvstore.dll

### DIFF
--- a/util.py
+++ b/util.py
@@ -516,7 +516,8 @@ def patch_libcuda() -> bool:
             return False
 
         log.info(f'Patched libcuda.so saved to: {patched_library}')
-        set_environment('LD_PRELOAD', patched_library)
+        protonmain.g_session.env['LD_LIBRARY_PATH'] = f'{os.path.dirname(patched_library)}:{protonmain.g_session.env["LD_LIBRARY_PATH"]}'
+
         return True
 
     except Exception as e:
--- a/gamefixes-umu/umu-starcitizen.py
+++ b/gamefixes-umu/umu-starcitizen.py
@@ -1,15 +1,25 @@
 """Game fix for Star Citizen"""
 
+import os
 from protonfixes import util
 
 
 def main() -> None:
     # patch libcuda to workaround crashes related to DLSS
     # See: https://github.com/jp7677/dxvk-nvapi/issues/174#issuecomment-2227462795
-    util.patch_libcuda()
+    patched = util.patch_libcuda()
+    if patched:
+        # Satisfy check for the existence of these dlls when trying to initialize ngx
+        # Copy an existing DLL to these three names
+        env_path = os.path.join(util.protonprefix(), 'drive_c', 'windows', 'system32')
+        dest_files = ['cryptbase.dll', 'devobj.dll', 'drvstore.dll']
+        for file in dest_files:
+            link_path = os.path.join(env_path, file)
+            if not os.path.isfile(link_path):
+                os.symlink('security.dll', link_path)
 
     # RSI Launcher depends on powershell
     util.protontricks('powershell')
 
     # RSI Launcher animation
-    util.winedll_override('libglesv2', 'builtin')
+    util.winedll_override('libglesv2', 'b')


util.py
```diff
@@ -516,7 +516,22 @@ def patch_libcuda() -> bool:
             return False
 
         log.info(f'Patched libcuda.so saved to: {patched_library}')
-        set_environment('LD_PRELOAD', patched_library)
+        set_environment('LD_LIBRARY_PATH', patched_library)
```
R1kaB3rN, line 519: We should never overwrite this environment variable in Proton. Only extend its value.

```
os.environ[‘LD_LIBRARY_PATH’] = f’{patched_library}:{os.environ[“LD_LIBRARY_PATH”]}’
```

R1kaB3rN (changes requested):

util.py
```diff
@@ -516,7 +516,22 @@ def patch_libcuda() -> bool:
...
+
+        # Satisfy games that check for the existence of these dlls when trying to initialize ngx
+        # Copy an existing DLL to these three names
+        env_path = os.path.join(protonprefix(), 'drive_c', 'windows', 'system32')
+        src_file = os.path.join(env_path, 'security.dll')
+
+        dest_file=os.path.join(env_path, 'cryptbase.dll')
+        shutil.copyfile(src_file, dest_file)
```
R1kaB3rN, line 527: Does creating a relative symlink not work?

R1kaB3rN: Also, whether a symlinks works or not, we should create the files in umu-starcitizen.py. Not this function.

mactan-sc: thanks for the feedback. I'll check in with the nvidia users if symlinks in this context are compatible with eac, either way will be incorporated into the gamefix instead of util

util.py
```diff
@@ -516,7 +516,8 @@ def patch_libcuda() -> bool:
             return False
 
         log.info(f'Patched libcuda.so saved to: {patched_library}')
-        set_environment('LD_PRELOAD', patched_library)
+        os.environ['LD_LIBRARY_PATH'] = f'{patched_library}:{os.environ["LD_LIBRARY_PATH"]}'
```
R1kaB3rN, line 519: After this line, we’ll also need update Proton’s environment. 

```
protonmain.g_session.env[‘LD_LIBRARY_PATH’] = os.environ[‘LD_LIBRARY_PATH’]

```

R1kaB3rN (changes requested):

util.py
R1kaB3rN: Edit:

```
protonmain.g_session.env[‘LD_LIBRARY_PATH’] = f'{patched_library}:{protonmain.g_session.env[“LD_LIBRARY_PATH”]}’
```

util.py
```diff
@@ -516,7 +516,9 @@ def patch_libcuda() -> bool:
...
```
R1kaB3rN, line 519: Since we’re using LD_LIBRARY_PATH instead of LD_PRELOAD, we’ll actually need to use the parent directory of the patched library, which is `config.path.cache_dir`. 

```
os.environ['LD_LIBRARY_PATH'] = f'{os.path.dirname(patched_library)}:{os.environ["LD_LIBRARY_PATH"]}'
```

util.py
```diff
@@ -516,7 +516,9 @@ def patch_libcuda() -> bool:
...
+        protonmain.g_session.env['LD_LIBRARY_PATH'] = os.environ['LD_LIBRARY_PATH']
```
R1kaB3rN, line 520:
```
protonmain.g_session.env[‘LD_LIBRARY_PATH’] = f'{os.path.dirname(patched_library)}:{protonmain.g_session.env[“LD_LIBRARY_PATH”]}’
```

R1kaB3rN (changes requested):

R1kaB3rN:
cc @ProjectSynchro for review

R1kaB3rN:
I’m also expecting the DLL override for this fix to not work:

util.winedll_override('libglesv2', 'builtin')

We’ll need to change ‘builtin’ to ‘b’.

ProjectSynchro:
No complaints or issues here :)

Your change requests look good to me

I agree with the change to `winedll_override` as well, looks like enforcing the types of overrides you can pass into that function has changed: 

https://github.com/Open-Wine-Components/umu-protonfixes/blob/d20e27768605afb6b0f9ed01e9fd2a0fba0d5d91/util.py#L423-L427

PR #301, opened 2025-04-29
Fix lotro store for standalone installer version

mdmatthias:
```
"""Disable libglesv2"""
    ## gpu acceleration on wined3d https://bugs.winehq.org/show_bug.cgi?id=44985
    # Make the store work.
    util.winedll_override('libglesv2', 'd')
```
This override actually breaks the store for the standalone lotro installer. When I open the store my fps drops and the store stays an empty black screen.
But when I run it without the libglesv2 override the store loads fine and no fps drop.
I'm not sure if this is the same for the steam version, so I moved it to a seperate umu fix instead of symlink and removed the libglesv2 override.

### DIFF
--- a/gamefixes-umu/umu-212500.py
+++ b/gamefixes-umu/umu-212500.py
@@ -1,1 +1,180 @@
-../gamefixes-steam/212500.py
+"""Game fix The Lord of the Rings Online"""
+
+from protonfixes import util
+import subprocess
+import sys
+from tempfile import mkstemp
+import __main__ as protonmain
+
+source = """
+def _mouse_fix_subprocess() -> None:
+    # this will only be imported in the temporary script
+    import struct
+    import time
+    import fcntl
+    from Xlib.X import LASTEvent
+    from Xlib.display import Display
+    from Xlib.ext import xinput
+    from Xlib.error import XError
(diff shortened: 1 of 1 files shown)


R1kaB3rN:
Sure, this is fine. Also, would you mind testing against Proton 10? https://github.com/ValveSoftware/Proton/releases/tag/proton-10.0-1b

mdmatthias:
@R1kaB3rN Looks like this is no longer needed in proton 10. In proton 10 it works as it should without enabling the fix. 
Also when enabling error logs ( by removing stderr=subprocess.DEVNULL in subprocess ), I see this:
`

  File "/tmp/tmpsmyp16e3.py", line 154, in <module>

    sys.exit(_mouse_fix_subprocess())

  File "/tmp/tmpsmyp16e3.py", line 11, in _mouse_fix_subprocess

    from Xlib.display import Display

  File "/home/matthias/.config/heroic/tools/proton/GE-Proton10-1/protonfixes/_vendor/Xlib/display.py", line 26, in <module>

    from six import create_unbound_method

ModuleNotFoundError: No module named 'six'
'
So my fix doesnt even work anymore in proton 10.

PR #303, opened 2025-05-01
fix: use correct values when overriding wine DLLs

R1kaB3rN:
A change to the `winedll_override` interface had invalidated a few fixes such as Star Citizen for a while. In the near future, we need to use a static type checker in our gamefixes to prevent cases like this.

See https://gitlab.winehq.org/wine/wine/-/wikis/Wine-User's-Guide#winedlloverrides-dll-overrides

### DIFF
--- a/gamefixes-steam/105000.py
+++ b/gamefixes-steam/105000.py
@@ -7,4 +7,4 @@ from protonfixes import util
 
 
 def main() -> None:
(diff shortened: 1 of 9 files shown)


Root-Core:
This is fixed in #201, which would also allow us to finally enable static type checking.

PR #307, opened 2025-05-02
gamefixes: Add Celeste fix for EGS build

termdisc:
Add `FNA3D_FORCE_DRIVER=OpenGL` to ensure that Vulkan is not used and that OpenGL is used.


termdisc:
User reported that the variable is not required any longer. Closing this for now.

PR #310, opened 2025-05-06
build: drop xrandr

R1kaB3rN:


util.py
```diff
@@ -822,10 +822,15 @@ def set_xml_options(
     return True
 
 
-def get_resolution() -> tuple[int, int]:
+def get_resolution() -> Union[tuple[int, int], None]:
```
Root-Core, line 825: The `Optional[]` type hint is preferred. We maybe should add a rule for that.

R1kaB3rN: As far as I can tell, ruff does not have one. Though, I haven't checked pyright or mypy yet.

Related to Optional, in ruff there's https://docs.astral.sh/ruff/rules/implicit-optional/#implicit-optional-ruf013 and https://docs.astral.sh/ruff/rules/non-pep604-annotation-optional/. We can add the former, but wouldn't enforce the favor of Optional over Union[..., None].

Root-Core: https://docs.astral.sh/ruff/rules/non-pep604-annotation-optional/ looks good to me, `type | None` might even be better than `Optional[]`. We would need to have `from __future__ import annotations`, but why not?

PR #311, opened 2025-05-06
Add Marvel Rivals (EGS) fixes

termdisc:
Replace target exe and set `SteamDeck=1`

The EGS version does not launch with the default .exe file or without setting `SteamDeck=1`.

This also coincides with an update to `umu-database.csv` at https://github.com/Open-Wine-Components/umu-database/pull/89

### DIFF
--- a/gamefixes-egs/umu-2767030.py
+++ b/gamefixes-egs/umu-2767030.py
@@ -1,0 +1,16 @@
+"""Game fix for Marvel Rivals"""
+# EGS-ID 575efd0b5dd54429b035ffc8fe2d36d0
+
+from protonfixes import util
+
+
+def main() -> None:
+
+    util.replace_command('epic_launch_helper.exe', 'MarvelRivals_Launcher.exe')
+
+    # Set SteamGameId so that non-steam versions can pick up steam-specific fixes in proton's wine code
+    util.set_environment('SteamGameId', '2767030')
+    
+    # Set SteamDeck=1 to be able to launch the game
+    util.set_environment('SteamDeck', '1')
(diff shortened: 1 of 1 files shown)


gamefixes-egs/umu-2767030.py
```diff
@@ -0,0 +1,16 @@
...
+
+from protonfixes import util
+
+
+def main() -> None:
+
+    util.replace_command('epic_launch_helper.exe', 'MarvelRivals_Launcher.exe')
+
+    # Set SteamGameId so that non-steam versions can pick up steam-specific fixes in proton's wine code
+    util.set_environment('SteamGameId', '2767030')
```
R1kaB3rN, line 12: umu-launcher automates setting this: https://github.com/Open-Wine-Components/umu-launcher/blob/main/umu/umu_run.py#L228

R1kaB3rN: Unless, like you had commented, this is added for users who will add the title as a non-Steam game and set GAMEID in the client’s launch options.

termdisc: I had taken inspiration from https://github.com/Open-Wine-Components/umu-protonfixes/blob/ee2927ea41e819817ae36f7231b5ba568db7c4e2/gamefixes-egs/umu-2009100.py#L8 in combination with recently seeing that Marvel Rivals got called out in the [Proton 10 notes](https://github.com/ValveSoftware/Proton/releases/tag/proton-10.0-1b).

I was not aware that the environment variable would be automated otherwise.

PR #316, opened 2025-05-10
add fixes for mashed

ChrisOboe:
This PR adds fixes for mashed. 
It will set wine to wine xp mode since this prevents that cars will fall through the ground in some stages. And it adds some launch args that will hide the launcher.

### DIFF
--- a/gamefixes-steam/281280.py
+++ b/gamefixes-steam/281280.py
@@ -1,0 +1,12 @@
+"""Game fix for Mashed(281280)"""
+
+from protonfixes import util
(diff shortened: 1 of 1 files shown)


R1kaB3rN:
Do you mind creating an [upstream report](https://github.com/ValveSoftware/Proton/issues) for this title using the latest Proton 10 beta?

ChrisOboe:
I created the upstream report here: https://github.com/ValveSoftware/Proton/issues/8683

PR #324, opened 2025-05-17
protonfixes: move more protonfixes related functions out of the proton script

loathingKernel:
I am not certain about these changes myself, it might have the opposite result, but the goal is to reduce the modifications required for any proton version to include umu and protonfixes.

In that regard, the modifications for running winetricks and setting up mountpoint drives have been moved into protonfixes. There is one more group of modifications that might be movable, the ones related to moving prefixes between steamdeck and normal modes.


loathingKernel:
Usage example
```python
    g_session.init_wine()

    import protonfixes

    if g_proton.missing_default_prefix():
        g_proton.make_default_prefix()

    g_session.init_session(sys.argv[1] != "runinprefix")

    protonfixes.setup(g_session.env, "PATH", ld_path_var, prepend_to_env_str)
    protonfixes.setup_mount_drives(setup_dir_drive)
    protonfixes.winetricks(g_session.env, g_proton.wine_bin, g_proton.wineserver_bin)
    protonfixes.execute()

    #determine mode
    rc = 0
    if sys.argv[1] == "run":
```

PR #327, opened 2025-05-19
Add Pokemon Gamma Emerald

BlueInterlude:
* Link: https://undreamedpanic.itch.io/gamma-emerald

Database PR: https://github.com/Open-Wine-Components/umu-database/pull/91 

See database PR first before merging.


R1kaB3rN:
If you're concerned whether it would be ok to add a fix for this, then it's fine. The only ones who should really be concerned is that platform and the creators of that game. In this case, umu-protonfixes is only running a winetricks verb when given that metadata. It would another thing though, if this module were to download that game.

gamefixes-umu/umu-pokemongammaemerald.py
```diff
@@ -0,0 +1,7 @@
+"""Game fix for Pokemon Gamma Emerald"""
+
+from protonfixes import util
+
+
+def main() -> None:
+    util.protontricks('vcrun2022')
```
R1kaB3rN, line 7: We don't need this file here because we have the folder `gamefixes-itchio`

R1kaB3rN (changes requested):

gamefixes-umu/umu-pokemongammaemerald.py
R1kaB3rN: Also, if you've observed that this is a frequent verb that needs to be installed for itchio games, then we can run the verb by default for that store front.

R1kaB3rN: For example, like how umu-protonfixes will run vcrun2022 for all EGS games:

https://github.com/Open-Wine-Components/umu-protonfixes/blob/4f229412a851bde3e330486e44051569eb5f0d4d/gamefixes-egs/default.py

BlueInterlude: Somehow missed the itchio folder, moved it there. 

I can double check and do some testing of my itch library this week to see if we should install this by default though.

gamefixes-itchio/umu-pokemongammaemerald.py
```diff
@@ -0,0 +1,7 @@
+"""Game fix for Pokemon Gamma Emerald"""
```
R1kaB3rN, line 1: Let's change the file name so the ID is added.

`gamefixes-itchio/umu-3534339.py`

R1kaB3rN: However, I believe there'll be some risk for ID collisions with Steam's. The integer can be of similar length to Steam's and is different than EGS's. When configured, umu-launcher will set SteamGameId=3534339 potentially applying an unrelated wine fix if it exists. At worst, the wrong fix can be applied for a game. Though, this isn't won't occur for this case: https://steamdb.info/app/3534339/.

R1kaB3rN:
LGTM, but would like some feedback from other maintainers before merging. For this, if we care about ID collisions, we'd need protonfixes to compute unique values if isn't already on Steam then assign new values to the relevant environment variables.

Linking database discussion: https://github.com/Open-Wine-Components/umu-database/pull/91

PR #334, opened 2025-05-29
util.patch_libcuda reapply read misplaced from refactor

mactan-sc:
Possibly due to 889a5b1 the read of binary changed to trying to read the destination instead of the source. resulting in:

```
ProtonFixes[72310] INFO: Found 64-bit libcuda.so at: /usr/lib/pressure-vessel/overrides/lib/x86_64-linux-gnu/libcuda.so.1
ProtonFixes[72310] CRIT: Unable to read libcuda.so: [Errno 2] No such file or directory: '/home/xxxx/.cache/protonfixes/libcuda.patched.so'
```

This returns the open() routine to reading the bytes from libcuda_path instead of the destination in .cache
cc  @ProjectSynchro


util.py
```diff
@@ -557,7 +557,8 @@ def patch_libcuda() -> bool:
 
         patched_library = config.path.cache_dir / 'libcuda.patched.so'
         try:
-            binary_data = patched_library.read_bytes()
+            with open(libcuda_path, 'rb') as f:
```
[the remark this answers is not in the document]
ProjectSynchro: You may want to ensure that libcuda_path is a path and not a string with: `binary_data = Path(libcuda_path).read_bytes()`

util.py
```diff
@@ -557,7 +557,7 @@ def patch_libcuda() -> bool:
 
         patched_library = config.path.cache_dir / 'libcuda.patched.so'
         try:
-            binary_data = patched_library.read_bytes()
+            binary_data = Path(libcuda_path).read_bytes()
```
R1kaB3rN, line 560: This fixes the issue.

However, after re-reviewing, it would be a bit better to write to actually write to a temporary directory instead of the cache. Neither this implementation nor the [starcitizen fix](https://github.com/Open-Wine-Components/umu-protonfixes/blob/00770111788d2049777d26143fd5466ee7fc69e4/gamefixes-umu/umu-starcitizen.py#L7) will check the cache to see if the patched file has already been created.

util.py
```diff
@@ -557,7 +557,7 @@ def patch_libcuda() -> bool:
 
         patched_library = config.path.cache_dir / 'libcuda.patched.so'
```
R1kaB3rN, line 558: Something like:

```
from tempfile import mkdtemp
patched_library =  f"{mkdtemp()}/libcuda.patched.so"
```

R1kaB3rN: However if the original intention of the fix was to refer to the cache, then cache should be checked for the patched file in umu-starcitizen.py.

R1kaB3rN:
Merging for now. This PR fixes the issue and either writing to a tmpfs or referring back to the cache for the patched file can be addressed separately.

util.py
ProjectSynchro: When I first wrote this fix the idea was to to clobber the library on every launch as it is grabbed from the nvidia driver. 

When the driver is updated the library we patched may not match the ABI that the driver expects and cause issues, granted creating a temporary directory also works as that will also be different on each launch.

PR #337, opened 2025-06-04
patch_libcuda: Fix patched library name and use a temporary directory instead of .cache

ProjectSynchro:
Making sure we use the right name should ensure that this library is loaded correctly as it's path is added to the beginning of LD_LIBRARY_PATH. 

Additionally using a tempdir makes sense as the cache that is currently being used isn't really a cache, it's just being used as a temporary directory anyways.

Let me know if you have any additional changes in mind.

### DIFF
--- a/util.py
+++ b/util.py
@@ -512,8 +512,6 @@ def patch_libcuda() -> bool:
 
     Returns true if the library was patched correctly. Otherwise returns false
     """
-    config.path.cache_dir.mkdir(parents=True, exist_ok=True)
-
     try:
         # Use shutil.which to find ldconfig binary
         ldconfig_path = shutil.which('ldconfig')
--- a/util.py
+++ b/util.py
@@ -586,10 +586,19 @@ def patch_libcuda() -> bool:
             return False
 
         log.info(f'Patched libcuda.so saved to: {patched_library}')
-        protonmain.g_session.env['LD_LIBRARY_PATH'] = (
-            f'{patched_library.parent}:{protonmain.g_session.env["LD_LIBRARY_PATH"]}'
-        )
 
+        # Set LD_LIBRARY_PATH and LD_PRELOAD to include the directory (or absolute path) of the patched library
+        current_ld_path = protonmain.g_session.env.get("LD_LIBRARY_PATH")
+        protonmain.g_session.env["LD_LIBRARY_PATH"] = (
+            f"{patched_library.parent}:{current_ld_path}" if current_ld_path else str(patched_library)
+        )
+        log.info(f'LD_LIBRARY_PATH updated to {protonmain.g_session.env.get("LD_LIBRARY_PATH")}')
+
+        current_ld_preload = protonmain.g_session.env.get("LD_PRELOAD")
+        protonmain.g_session.env["LD_PRELOAD"] = (
+            f"{patched_library}:{current_ld_preload}" if current_ld_preload else str(patched_library)
+        )
+        log.info(f'LD_PRELOAD updated to {protonmain.g_session.env.get("LD_PRELOAD")}')
         return True
 
     except Exception as e:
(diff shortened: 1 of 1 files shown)


ProjectSynchro:
I'll mark this as ready after some testing is done just to ensure that the changes are sticking correctly.

ProjectSynchro:
Added the option to set LD_PRELOAD as well, both that and LD_LIBRARY_PATH will append the new paths if the environment variable is already set, or it will just set it outright if that is not the case.

ProjectSynchro:
This has been tested and confirmed working by some of the folks in the Star Citizen Linux User Group. Marking this as ready for review.

util.py
```diff
@@ -588,10 +586,19 @@ def patch_libcuda() -> bool:
             return False
 
         log.info(f'Patched libcuda.so saved to: {patched_library}')
-        protonmain.g_session.env['LD_LIBRARY_PATH'] = (
-            f'{patched_library.parent}:{protonmain.g_session.env["LD_LIBRARY_PATH"]}'
+
+        # Set LD_LIBRARY_PATH and LD_PRELOAD to include the directory (or absolute path) of the patched library
+        current_ld_path = protonmain.g_session.env.get("LD_LIBRARY_PATH")
+        protonmain.g_session.env["LD_LIBRARY_PATH"] = (
+            f"{patched_library.parent}:{current_ld_path}" if current_ld_path else str(patched_library)
```
R1kaB3rN, line 593: When setting the new LD_LIBRARY_PATH, we can also refactor these few lines as:

```
ld_paths = [patched_library.parent, protonmain.g_session.env.get("LD_LIBRARY_PATH")]
protonmain.g_session.env["LD_LIBRARY_PATH"] = ":".join(filter(None, ld_paths))
```

util.py
```diff
@@ -588,10 +586,19 @@ def patch_libcuda() -> bool:
...
         )
+        log.info(f'LD_LIBRARY_PATH updated to {protonmain.g_session.env.get("LD_LIBRARY_PATH")}')
 
+        current_ld_preload = protonmain.g_session.env.get("LD_PRELOAD")
+        protonmain.g_session.env["LD_PRELOAD"] = (
+            f"{patched_library}:{current_ld_preload}" if current_ld_preload else str(patched_library)
```
R1kaB3rN, line 599:
```
ld_paths = [patched_library.parent, protonmain.g_session.env.get("LD_PRELOAD")]
protonmain.g_session.env["LD_PRELOAD"] = ":".join(filter(None, ld_paths))
```

R1kaB3rN: The current implementation is fine though.

Also, why is LD_PRELOAD being set again for this title? [According to a previous pull request](https://github.com/Open-Wine-Components/umu-protonfixes/pull/294), the anti-cheat for the game would clear the value.

util.py
ProjectSynchro: Implemented in https://github.com/Open-Wine-Components/umu-protonfixes/pull/337/commits/34b9208e3e8426fed2b1a62b362a941a4be47b0f

util.py
ProjectSynchro: Implemented in https://github.com/Open-Wine-Components/umu-protonfixes/pull/337/commits/34b9208e3e8426fed2b1a62b362a941a4be47b0f

ProjectSynchro: > No, that's fine. After all, you've already confirmed this fix has been tested and has worked when setting both LD_PRELOAD and LD_LIBRARY_PATH.
> 
> > Some users were having more luck setting that instead of using LD_LIBRARY_PATH for some reason
> 
> Yeah, that's strange and doesn't seem right...

Yeah.. I suspect something else has gone wrong in their environments or something.. Anyways this at least makes doubly sure that the library we are overriding is being loaded (Setting too much is better than not setting enough :smile: )

PR #359, opened 2025-07-06
Workaround to fix Sifu freezing

UsernamesAreNotMyThing:
Apparently, the game's long-standing freezing problems under Proton are fixable through editing a config file, but the file is in the `AppData/Local` folder in the prefix, so I also had to update `_get_config_full_path` and `BasePath` because that directory is not reachable with either of the two options available, and may also come in handy in case there are other games out there that put their config files in the `AppData` folder.

https://github.com/ValveSoftware/Proton/issues/6657#issuecomment-3037914449

### DIFF
--- a/gamefixes-steam/2138710.py
+++ b/gamefixes-steam/2138710.py
@@ -1,8 +1,8 @@
 """Sifu"""
 
-from protonfixes import util, BasePath
+from protonfixes import util
 
 
 def main() -> None:
     """Stops the game from freezing."""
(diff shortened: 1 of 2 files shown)


UsernamesAreNotMyThing:
I cannot test it myself, I do not have the game.

GloriousEggroll:
doesnt seem like a significantly bad change or anything that would affect other games for now, will go ahead and merge. we can always revert it

PR #370, opened 2025-08-07
utilities: move proton utility functions into their own file

loathingKernel:
add functions to download the required dlls to upgrade DLSS, XeSS and FSR4


loathingKernel:
This is ready for review. 

* Wine patch: [0001-HACK-kernelbase-allow-overriding-dlls-for-DLSS-XeSS-.patch](https://github.com/user-attachments/files/22707291/0001-HACK-kernelbase-allow-overriding-dlls-for-DLSS-XeSS-.patch)

* Proton patch: I'll add it later if this gets into protonfixes, it's too different right now.

loathingKernel:
> It may be called from protonfixes later, via a game specific config file. Or would it be better to set an ENV var in this case?

These are not game-specific, they are settings that custom protons can provide though. Currently enabling them is controlled through proton options, and some of them require access to proton script internals such as the the path to the compatdata or the environment. I like the idea of protonfixes handling them in their entirety without needing custom bits inside the proton script if it is possible.

I think that upscalers can go into their own file indeed, for the rest I am not sure yet either. There's another thing that I wanted to [move from proton into protonfixes](https://github.com/GloriousEggroll/proton-ge-custom/blob/master/proton#L62-L159) at some point but never got around to it.

PR #384, opened 2025-08-26
Add d3dx9 for Two Worlds: Epic Edition

marciocr:
The game without this dependency has a bug where the image starts to flicker as if it were a rainbow. The fix is ​​documented in protondb.


R1kaB3rN:
https://github.com/ValveSoftware/Proton/issues/5454

PR #402, opened 2025-09-30
Added offline support for winetricks

sheikhmishar:
After initial installation, internet is not necessary for tweaking stuff, although adding packages will require internet. So, I removed the mandate of being online

### DIFF
--- a/util.py
+++ b/util.py
@@ -310,8 +310,7 @@ def protontricks(verb: str) -> bool:
             # Proceed with your function logic here
             pass
         else:
-            log.info('No internet connection. Winetricks will be skipped.')
-            return False
+            log.info('No internet connection. Winetricks will not be able to download anything.')
 
         log.info('Installing winetricks ' + verb)
         env = dict(protonmain.g_session.env)


util.py
```diff
@@ -310,8 +310,7 @@ def protontricks(verb: str) -> bool:
             # Proceed with your function logic here
             pass
         else:
-            log.info('No internet connection. Winetricks will be skipped.')
-            return False
+            log.info('No internet connection. Winetricks will not be able to download anything.')
```
R1kaB3rN, line 313: Mind clarifying a bit more? The way I see it, this change will detriment the UX for clients like Heroic Games Launcher or when running through Steam by slowing down launch times when the network link suddenly goes down.

By not returning when the system's network links are down, winetricks will attempt to fetch the verb's dependencies without checking its cache and will hang for a period of time. Specifically, [winetricks@e73c4d8](https://github.com/Winetricks/winetricks/tree/e73c4d8f71801fe842c0276b603d9c8024d6d957) invokes a `WINETRICKS_DOWNLOADER` and configures it with `WINETRICKS_DOWNLOADER_RETRIES` and `WINETRICKS_DOWNLOADER_TIMEOUT`. Depending on the verb, the hang time period can be a very long (45s+ for each HTTP request by default) for the user. 45s+ for each request isn't unacceptable, and what makes it worse is that most clients don't track or have control over that download state when offline. They're just currently trusting that all the downloads will complete, and that the protonfix will be applied fully. Also, this code path is mostly applicable for verbs rather than tweaks. While there are some titles like [gamefixes-steam/33460.py](https://github.com/Open-Wine-Components/umu-protonfixes/blob/487bca55aa9abb3f93d10a5169c4ba3f53a69160/gamefixes-steam/33460.py) that don't make network requests, protonfixes won't hit this code path, and we don't really have any fixes that purely tweak stuff through winetricks. So I'm not really seeing how this change can be really useful for most users. If anything, it's more detrimental.

For offline support here, you need to determine if something like a setting, which requires no network request, is being applied or not. Otherwise, for verbs, winetricks needs to change to fallback to its cache when installing verbs when the network link is down.

R1kaB3rN:
Closing. Will consider reopening on clarification.

PR #416, opened 2025-10-22
Space Engineers - dotnet48

erlking:
Hi,
as dotnet48 is currently working could you add it to the gamefixes for Space Engineers?
without the performance is terrible and multiplayer (with mods) barely possible at all.

thanks!

### DIFF
--- a/gamefixes-steam/244850.py
+++ b/gamefixes-steam/244850.py
@@ -5,7 +5,6 @@ from protonfixes import util
 
 def main() -> None:
     util.protontricks('xaudio29')
-    util.protontricks('vcrun2022')
     util.protontricks('dotnet48')
 
     base_attibutte = '<runtime>'


gamefixes-steam/244850.py
```diff
@@ -5,6 +5,7 @@
 
 def main() -> None:
     util.protontricks('xaudio29')
+    util.protontricks('dotnet48')
```
R1kaB3rN: > Hi, as dotnet48 is currently working could you add it to the gamefixes for Space Engineers? without the performance is terrible and multiplayer (with mods) barely possible at all.
> 
> thanks!

Since the game works without the verb and you're claiming that installing it improves performance, do you mind providing some benchmarks?

I would be open to including this, but, generally, we only create fixes for titles that are broken in some way.

R1kaB3rN (approved):

GloriousEggroll:
winetricks dotnet48 appears to break this option:
`<gcServer enabled = "true" />`:

https://github.com/GloriousEggroll/proton-ge-custom/issues/333

need to choose one or the other either dotnet48 or no dotnet and `<gcServer enabled = "true" />`.

anyone able to confirm which delivers better performance?

PR #426, opened 2025-11-02
Add fix for The Journeyman Project 3: Legacy of Time

CommandMC:
This game needs the D: drive pointing to `<game install dir>/data`, otherwise it won't launch

### DIFF
--- a/gamefixes-gog/umu-1207659009.py
+++ b/gamefixes-gog/umu-1207659009.py
@@ -1,0 +1,7 @@
+"""Fix for The Journeyman Project 3: Legacy of Time"""
+
+from protonfixes import util
+
+
+def main() -> None:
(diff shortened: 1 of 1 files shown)


gamefixes-gog/umu-1207659009.py
```diff
@@ -0,0 +1,10 @@
+"""Fix for The Journeyman Project 3: Legacy of Time"""
+
+from pathlib import Path
+
+from protonfixes import util
+
+
+def main() -> None:
+    game_dir = Path(util.get_game_install_path()) / 'data'
+    util.create_dos_device('d', path=str(game_dir))
```
R1kaB3rN: Nit: 

Would prefer:
```
util.create_dos_device('d', path=f'{util.get_game_install_path()}/data')
```
But no biggie.

Issue #430, opened 2025-11-03
Add fix for the Stove Launcher (Non-Steam)


BlueInterlude:
The Stove launcher (https://www.onstove.com/en/download) requires Webview2 to run properly. See the Lutris script for reference, https://lutris.net/games/install/38109/view. You can install WebView2 here, https://developer.microsoft.com/en-us/microsoft-edge/webview2/?form=MA13LH#download. 

There's a couple of open issues on the Winetricks repository to add support for WebView2 and seems like the main developer of Winetricks is willing to accept PRs for it but that's above my skill level. See https://github.com/Winetricks/winetricks/issues/1986 for the issue with the most comments and see https://github.com/Winetricks/winetricks/issues?q=WebView2 for the other related issues. 

Seems like this could be covered by umu manually for now until Winetricks adds support for it. 

I'd PR this myself but I'm not sure how to add a proton fix to install a dependency not covered by Winetricks. 

Thank you:)

R1kaB3rN:
Closing as won't fix. See https://github.com/Open-Wine-Components/umu-protonfixes/pull/431#issuecomment-3593514850.

Feel free to open another pull request if you had discovered another method of fixing the launcher. I think adding STOVE Launcher as a storefront would be a nice addition to this project, so if it would be possible to fix, that would be great.

As a suggestion, you can also try contacting STOVE Launcher support about considering their developers to bundle a Fixed version of the webview2 runtime to support those still on Windows 7, which they claim on their website to support. Though, I would not get any hopes up on that.

PR #431, opened 2025-11-05
Add fix for STOVE Launcher

BlueInterlude:
Database PR: https://github.com/Open-Wine-Components/umu-database/pull/109

Issue mentioned here https://github.com/Open-Wine-Components/umu-protonfixes/issues/430#issuecomment-3488600349 is fixed by using `PROTON_ENABLE_WAYLAND` but I'm not sure if that should be set as part of a proton fix. Without it, the launcher is still usable albeit glitchy.

### DIFF
--- a/verbs/webview2.verb
+++ b/verbs/webview2.verb
@@ -9,10 +9,10 @@ load_webview2()
 {
     # https://developer.microsoft.com/en-us/microsoft-edge/webview2/
     if [ "${W_ARCH}" = "win64" ]; then
-        w_download https://msedge.sf.dl.delivery.mp.microsoft.com/filestreamingservice/files/2e372271-55ba-41c0-89b9-dfc306dee437/MicrosoftEdgeWebView2RuntimeInstallerX64.exe
+        w_download https://msedge.sf.dl.delivery.mp.microsoft.com/filestreamingservice/files/b440317f-62d7-4af2-9099-8f3abdc40ae0/MicrosoftEdgeWebView2RuntimeInstallerX64.exe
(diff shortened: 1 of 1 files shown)


gamefixes-umu/umu-stovelauncher.py
```diff
@@ -0,0 +1,9 @@
+"""STOVE Launcher"""
```
R1kaB3rN: Let's update the `webview2` verb a bit by making the installation silent.

Have you already found the installer's command line options for that? If not and it isn't possible to suppress the installer window, we may need to find another installer that supports unattended installation.

R1kaB3rN: Adding `/silent` and `/install` should do it. Just be sure those options are only applied when configured as unattended.

BlueInterlude: That works pretty well, thank you! Added it here:)

BlueInterlude: Alright that should be all set now.

R1kaB3rN (approved):

GloriousEggroll:
unfortunately we can't distribute webview2, per:

https://github.com/ValveSoftware/Proton/issues/8448#issuecomment-2676457414
https://github.com/ValveSoftware/Proton/issues/8448#issuecomment-2676461114
https://github.com/ValveSoftware/Proton/issues/8448#issuecomment-2676654674

specific details are in the last comment, youll have to expand it as kisak decided to hide it as off-topic.

PR #437, opened 2025-11-11
Add fix for Not For Broadcast

Tiagoquix:
This game is heavily reliant on video playback, so much so that it doesn't work with default Proton. Proton-GE is needed.

There are some beta branches of the game that aim to fix video playback that some users experience. I've tried only one of them, but it didn't help me.

Then I remembered about the GST media thing and I tried the normal version (without beta branches selected) and in comparison to winedmo backend it seems all video/game freezing and crashing is fixed.

https://store.steampowered.com/app/1147550/Not_For_Broadcast/

This fix was tested by completing the first chapter of the game successfully. Without this fix, the game crashes/freezes entirely after less than 5 minutes of gameplay (even in the main menu).

Tested with GE-Proton10-25.

Database PR: https://github.com/Open-Wine-Components/umu-database/pull/115

### DIFF
--- a/gamefixes-steam/1147550.py
+++ b/gamefixes-steam/1147550.py
@@ -1,0 +1,9 @@
+"""Game fix for Not For Broadcast"""
+
+from protonfixes import util
+
+
+def main() -> None:
+    # Force using winegstreamer instead of winedmo for video playback
+    # Fixes video/game freezing and crashing
+    util.set_environment('PROTON_MEDIA_USE_GST', '1')


R1kaB3rN:
We can also create a fix for the game’s prologue as well. 

https://github.com/ValveSoftware/Proton/issues/3355

Tiagoquix:
@R1kaB3rN I have not played the Prologue version, but I suppose the same technology is present there, so it's fine for me to add for it too. Would you like me to do a PR for it?

In regards to the posts: yeah, the company does not support Linux. It's a 50/50 wherever it will work on Linux on not. From my testing of the Alternate Video beta branch, the game would not disappear and KDE's taskbar (task manager) would be visible at all times at the top of the game. So there's also problems with that.

The second topic you've linked doesn't seem to mention Linux specifically, so I think it's fine from our side.

PR #451, opened 2025-11-23
Add fix for Frame Generation in Black Myth: Wukong

loathingKernel:
* dxvk-nvapi docs: https://github.com/jp7677/dxvk-nvapi/wiki/Tips-and-tricks-for-usage-with-DXVK-NVAPI#dlss-frame-generation

Draft while I am waiting on some people to test.

### DIFF
--- a/gamefixes-steam/2358720.py
+++ b/gamefixes-steam/2358720.py
@@ -3,6 +3,11 @@
 from protonfixes import util
 
 def main() -> None:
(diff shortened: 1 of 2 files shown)


loathingKernel:
Fix seems to disable FSR when enabled, so it has to be enabled per-gpu. Not sure if that's possible through protonfixes presently. I'll see if there's an easy way to do it.

PR #462, opened 2025-12-08
fix: plants vs zombies goty slowdown

xXJSONDeruloXx:
simple fix to pvz goty which monitors prefix for the good exe which only gets created after game is running, then backs up original exe and coppies good one into game file. Also relaunches to good exe if not yet swapped over.

pretty much just scripting out the fix described here and cited in a lot of protondb reports: https://www.youtube.com/watch?v=EK1HyVGlHvY

lmk if theres a better or simpler way!

### DIFF
--- a/gamefixes-steam/3590.py
+++ b/gamefixes-steam/3590.py
@@ -18,14 +18,13 @@ from protonfixes.logger import log
 
 
 def _fix_executable() -> None:
-    """Wait for correct exe in prefix, then backup and replace game exe"""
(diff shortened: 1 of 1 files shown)


GloriousEggroll:
you have an error due to invalid logger call.

It should be:

```
from protonfixes.logger import log

log.warn
```

there is no log.error

xXJSONDeruloXx:
@GloriousEggroll my bad, updated

GloriousEggroll:
is this good to go? will merge if so

PR #471, opened 2025-12-16
adapt Tarkov fixes for outside of steam

Jan200101:
- add a file to check if service was already configured and configures it if its missing
    - there are setups where the BattlEye service installs but the service is only configured to start manually so it never runs and the Launcher complains. 
- checks if service files were already installed
- looks for BattlEye files from the Steam version and BsgLauncher version
- add umu symlink to steam fix

I added logging similar to the wintricks stuff to the service configuration so that the user is aware of whats happening
```
ProtonFixes[97958] INFO: Using global defaults for UNKNOWN (3932890)
ProtonFixes[97958] INFO: Using local protonfix for UNKNOWN (3932890)
ProtonFixes[97958] INFO: Checking if winetricks "dotnet48" is installed
ProtonFixes[97958] INFO: Checking if winetricks "vcrun2022" is installed
ProtonFixes[97958] INFO: Checking if winetricks "dotnetdesktop6" is installed
ProtonFixes[97958] INFO: Checking if winetricks "dotnetdesktop8" is installed
ProtonFixes[97958] INFO: Checking if BattlEye Service is installed
```

Tested this locally using umu-launcher but I still needed to set PROTON_BATTLEYE_RUNTIME, wouldn't be an issue on Lutris and co.

### DIFF
--- a/gamefixes-steam/3932890.py
+++ b/gamefixes-steam/3932890.py
@@ -3,9 +3,12 @@ Install BattlEye Service - does not permit online play
 """
 
 import os
-import glob
 import shutil
+from pathlib import Path
 from protonfixes import util
+from protonfixes.logger import log
+
+SERVICE_FLAG = "service_configured"
 
 def main() -> None:
     util.install_battleye_runtime()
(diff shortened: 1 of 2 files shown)


Jan200101:
closing because with the current setup its fundementally impossible to have this fix be shared across steam and umu

PR #473, opened 2025-12-18
umu-starcitizen gamefixes

mactan-sc:
- Remove the libglesv2 override
  - The launcher animation lag no longer present on systems surveyed. The override has also conflicted with winewayland mode via PROTON_ENABLE_WAYLAND=1 by preventing the launcher window from displaying
- Remove ntsync override
  - The noted EAC conflict has not been observed for some time by players that have ntsync enabled
- Disable dxwebsetup.exe and dotNetFx45_Full_setup.exe
  - Prevent an infrequent but perpetual hang when the RSI  Launcher is installing or updating. Some star citizen linux community tools [LUG Helper](https://github.com/starcitizen-lug/lug-helper/blob/main/lug-helper.sh#L2205) and [rsi launcher flatpak](https://github.com/mactan-sc/rsilauncher/blob/main/scripts/rsi-run.sh#L42) each disable these

### DIFF
--- a/gamefixes-umu/umu-starcitizen.py
+++ b/gamefixes-umu/umu-starcitizen.py
@@ -21,8 +21,6 @@ def main() -> None:
     # RSI Launcher depends on powershell
     util.protontricks('powershell')
 
-    # RSI Launcher animation
-    util.winedll_override('libglesv2', util.OverrideOrder.BUILTIN)
-
-    # SC's shipped EAC Installer fails with ntsync
-    util.disable_ntsync()
+    # Prevent RSI Launcher install and update from hanging indefinitely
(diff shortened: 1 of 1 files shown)


gamefixes-umu/umu-starcitizen.py
```diff
@@ -21,8 +21,6 @@ def main() -> None:
     # RSI Launcher depends on powershell
     util.protontricks('powershell')
 
-    # RSI Launcher animation
-    util.winedll_override('libglesv2', util.OverrideOrder.BUILTIN)
```
R1kaB3rN: Do you know if the override is needed for those on XWayland/X11?

PR #519, opened 2026-03-22
feat: optiscaler as a launch arg

xXJSONDeruloXx:
Opening this draft PR mostly in interest of starting the discussion around adding a `PROTON_OPTISCALER` var in similar vein to the `PROTON_FSR4_UPGRADE` and similar vars in upscalers.py.

As discussed previously in: https://github.com/Open-Wine-Components/umu-protonfixes/pull/461

There's been expressed interest to implement a "one click" optiscaler implementation, however a major blocker has been the fact that the default usage of Optiscaler involves dropping many files into the game's executable directory, often with overwriting or renaming a key DLL.

What I have here is a MVP for adding opti to a given game WITHOUT touching any of the game's files or modifying the game directory itself. Essentially we download optiscaler from official distribution and extract to compatdata (or optionally user provides a path/to/opti for the newer previews not yet on github).

User provides one of the various DLL proxies as the value to the optiscaler var, for example `PROTON_OPTISCALER=dxgi`, at which point on game launch the optiscaler payload is either downloaded, or used from cache if exists, or pulled from valid path/to/dir if provided and those files are put into users prefix, NOT the game directory. part of this includes backing up via rename any dll named as the proxy dll is named in system32 and noting which proxy name used in manifest.json for janitor pass under removal conditions.

additionally, appends (proxy)n,b to WINEDLLOVERRIDES to facilitate the dll loading.

If user launches game without the optiscaler args, manifest and janitorial routines should clear the prefix back to prev state, which mirrors the user experience of removing the fsr upgrade arg as upscalers.py facilitates. 

I have tested this with a handful of games, and I am quite happy with the initial results. I also had previous drafts of this with other env vars exposed like `category.key=val;` for setting OptiScaler.ini values in launch args, but that seems a bit high touch currently and I solely want to start a discussion here and decide best path for this implementation, if there is in fact any. 

I think the added tests are a pretty readable way to see what the intended UX is here, I am sure there are plenty of corner cases not currently handled or considered.

Please let me know what you think about the current implementation and if it jives with the current conventions!

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -7,6 +7,7 @@ from typing import Callable
 
 from . import fix
 from .logger import log
+from .optiscaler import setup_optiscaler
 from .upscalers import setup_upscalers
 from .utilities import (
     setup_frame_rate,
@@ -100,6 +101,11 @@ def execute() -> None:
             if isinstance(dialog, ZenityWaitDialog):
                 dialog.start()
             fix.main()
+            setup_optiscaler(
+                os.environ,
+                os.environ['STEAM_COMPAT_DATA_PATH'],
+                os.path.join(os.environ['STEAM_COMPAT_DATA_PATH'], 'pfx'),
+            )
 
         except Exception:
             sys.stderr.write('ProtonFixes ' + traceback.format_exc())
@@ -116,6 +122,7 @@ __all__ = [
     'setup_frame_rate',
     'setup_local_shader_cache',
     'setup_mount_drives',
+    'setup_optiscaler',
     'setup_upscalers',
     'winetricks',
 ]
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ to stock behavior.
 PROTON_OPTISCALER=winmm %command%
 ```
 
-Optional variables:
+By default this uses the pinned upstream OptiScaler `0.7.9` release. Optional variables:
 
 ```bash
 PROTON_OPTISCALER_CONFIG=Menu.Scale=1.2;FSR.Fsr4ForceCapable=true
--- a/README.md
+++ b/README.md
@@ -25,8 +25,9 @@ You can also set [dxvk options](https://github.com/doitsujin/dxvk/wiki/Configura
 ```
 
 OptiScaler uses Proton-style environment variables instead of `-pf_` aliases. Set
-`PROTON_OPTISCALER` to a proxy like `auto`, `winmm`, `dxgi`, or `version`. If you remove the
-variable from your launch options, the next run restores the prefix to stock behavior.
+`PROTON_OPTISCALER` to a proxy like `auto`, `winmm`, `dxgi`, or `version` (or `1` as shorthand
+for `auto`). If you remove the variable from your launch options, the next run restores the prefix
+to stock behavior.
 
 ```bash
 PROTON_OPTISCALER=winmm %command%
--- a/README.md
+++ b/README.md
@@ -24,16 +24,20 @@ You can also set [dxvk options](https://github.com/doitsujin/dxvk/wiki/Configura
 -pf_dxvk_set=dxgi.maxFrameRate=40 -pf_dxvk_set=d3d9.maxFrameRate=40
 ```
 
-OptiScaler can also be staged into the Wine prefix without modifying the game directory.
-The first run should point `-pf_optiscaler` to an extracted OptiScaler payload directory, a `.zip`,
-or a `.7z` archive (requires `7z`/`7zz` in `PATH`). Later runs can reuse the staged payload with
-`-pf_optiscaler=1`, and `-pf_optiscaler=0` restores the original proxy DLL.
+OptiScaler uses Proton-style environment variables instead of `-pf_` aliases. Set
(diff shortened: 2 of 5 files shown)


loathingKernel:
Hi, thank you for the effort put into the PR. I saw your PR this morning, it's a lot of changes and it will take a bit to review. I just wanted to say that it's not ignored.

__init__.py
```diff
@@ -100,6 +101,11 @@ def execute() -> None:
             if isinstance(dialog, ZenityWaitDialog):
                 dialog.start()
             fix.main()
+            setup_optiscaler(
+                os.environ,
+                os.environ['STEAM_COMPAT_DATA_PATH'],
+                os.path.join(os.environ['STEAM_COMPAT_DATA_PATH'], 'pfx'),
+            )
```
loathingKernel, line 108: So far my approach has been to write such utilities in a way that each proton that wants to implement `umu-protonfixes` can decide whether they want to support the extra utilities other than the game fixes. That is why things like `setup_upscalers`, `setup_mount_drives` etc are only exposed in the module but never called by `umu-protonfixes` themselves.

xXJSONDeruloXx:
@loathingKernel totally understand, take your time. I wanted to mostly just open the discussion and see what the best approach looks like. Looking forward to your feedback!

abdullsuhbhan:
So this works without dropping or modifying original game files, right? If so then does it lessen the fear of banning and detection by Anti-Cheats?

loathingKernel:
> If so then does it lessen the fear of banning and detection by Anti-Cheats?

No, not in any way, shape or form.

xXJSONDeruloXx:
> So this works without dropping or modifying original game files, right? If so then does it lessen the fear of banning and detection by Anti-Cheats?

No, games will still "see" tampering as the DLL links to unplanned locations so there's no benefit there. The advantage of this approach for something like a proton launch arg is that it's non destructive to the game contents and mostly to the prefix as well, so recovering to stock is more painless

loathingKernel:
I've not forgotten about this, I have just been busy, and I wanted to test an idea that might make this even cleaner. I have expanded on the same idea that we use to
redirect `LoadLibraryExW` to load updated upscaler DLLs to make it configurable at runtime. Without getting into too much detail right now, things like this are now possible
```
WINE_LOADDLL_REDIRECT="nvngx_dlss.dll=c:\windows\system32\dlss\nvngx_dlss.dll"
```
this instructs wine to load a different dll from an alternative path when it tries to load `nvngx_dlss.dll` instead of the one passed to `LoadLibraryExW`.
This can also do the following kind of "rename", i.e. load a differently named DLL instead of the one it replaces
```
WINE_LOADDLL_REDIRECT="winmm.dll=c:\optiscaler\optiscaler.dll"
```
which theoretically could allow for optiscaler to be kept in a different directory without the need for cleanup. At this moment I am not sure if this works, there are a few caveats I have to investigate still for this use-case. since it is meant to be a shim for another DLL, and my approach at this point completely masks the wine built-in dll.

xXJSONDeruloXx:
That sounds interesting, I suppose in principle this would also still allow optiscaler dll to properly reference its surroundings in its own directory divorced from the game folder. Will mess with this concept and see how feasible this is

loathingKernel:
I did some testing and only `dbghelp.dll` offers a viable target for something like that presently, without much success in the couple of games that I tried it with (one needs it as `dxgi.dll` to work, the other one generally doesn't work). I still have to look into this further. On the bright side, Optiscaler did load in both games in my tests and presented the in-game gui.

loathingKernel:
I have been in contact with the OptiScaler developers while trying to figure out how to achieve this. After fixing a few cases where the configuration was not applied correctly, I believe I have a solid plan on how to implement this. I have tested this approach locally, which includes some modifications to wine and a pre-release build of OptiScaler. The plan consists of the following rough steps.

* Create and publish an OptiScaler archive through the manifest we use for dll downloads. This archive will contain the following files `OptiScaler.dll`, `OptiScaler.ini`, `amd_fidelityfx_loader_dx12.dll`, `amd_fidelityfx_upscaler_dx12.dll`, `amd_fidelityfx_framegeneration_dx12.dll`. The other upscaler dlls can be fetched through the current implementation.
* Extract the archive to `c:\windows\system32\umu\` and update the required paths in the config to point to the dlls we want to use.
* Create symlinks for the proxy dlls to `OptiScaler.dll`
* Set the expected proxy name through `WINE_OPTISCALER_NAME` env variable, i.e. `WINE_OPTISCALER_NAME=d3d12.dll`
* Do any other required configuration.

The reason I don't want to use the OptiScaler archive from GitHub releases directly is because we would have to ship `7z` with protonfixes, and to exclude files that we can obtain in different means.

The other thing I am uncertain of, is if we can obtain the `amd_fidelityfx_*` dlls from somewhere else and add them to the upscaler manifest as individual downloads.

xXJSONDeruloXx:
ayy, is this related? great to see! https://www.gamingonlinux.com/2026/05/proton-cachyos-11-adds-initial-optiscaler-integration-and-lots-of-other-fixes/

loathingKernel:
Ah, sorry, forgot to reply to this. Yes, it's related. Still needs refinement.

PR #526, opened 2026-03-24
chore: cleanup some fsr4 logic

powerofthe69:
The existing `or` command to check for a cached FSR 4 file was causing unnecessary network requests by checking the URLs before checking the cached files (because or statements resolve from left to right). This reverses that logic to check cached files first and avoids unnecessary network requests when launching games.

This also changes the default version to be the last entry in the dict. Up til now, this has been a hardcoded value that needs updated to prefer new releases. If the preference is to have the default version hardcoded as it was, then this change can be dropped.

### DIFF
--- a/upscalers.py
+++ b/upscalers.py
@@ -164,13 +164,13 @@ def __get_fsr4_dlls(version: str = 'default') -> dict:
         return cached_file.exists()
 
     if version == 'default' or version not in __fsr4_dlls.keys():
-        version = '4.0.3'
+        version = next(reversed(__fsr4_dlls))
 
     item = __fsr4_dlls[version]
-    if not (__dll_download_exists(item['download_url']) or _cached_file_exists(item)):
+    if not (_cached_file_exists(item) or __dll_download_exists(item['download_url'])):
         for key in sorted(__fsr4_dlls.keys(), reverse=True):
             item = __fsr4_dlls[key]
-            if __dll_download_exists(item['download_url']) or _cached_file_exists(item):
+            if _cached_file_exists(item) or __dll_download_exists(item['download_url']):
                 version = key
                 break
 


loathingKernel:
~~I have explained this, this was on purpose in case the dlls become available again. https://github.com/Open-Wine-Components/umu-protonfixes/issues/515#issuecomment-4099227806~~

Nvm, I thought it was removing the network check completely. On second look this looks fine.

Not being the 4.1.0 version was also on purpose, because I encountered comments on discord that it was causing broken graphics.

loathingKernel:
Ok, I remembered why I ordered the checks this way (yes, the way they are ordered is intentional exactly because of how `or` is evaluated). This is a version availability check, code after this point assumes that it will be able to find a version to download/use. If for whatever reason the cache is deemed to be untrustworthy, i.e. by a future checksum check, the result of this function needs to be a version that can be downloaded so it cannot depend predominantly on the existence of a cached file. It also logs the network failure before continuing. The cache check is a quickly added convenience to resolve the current situation of the missing dlls.

So yeah, do with this information as you will. I think this change should be merged, the situation calls for it, but I might need to find this information in the future in case I forget, so this is as good a place as any.

I should stop doing PR reviews before coffee, that's my bad.

powerofthe69:
> I thought it was removing the network check completely

Definitely didn't want to remove the network check completely! I think even if the files become available again, this is a nice to have in case they're ever delisted again too.

> I encountered comments on discord that it was causing broken graphics

I wonder if this is due to the DLL that was being used being a leaked version, rather than the one officially released a week later. There is a size difference of ~0.3 MB. I haven't noticed any issues with this official one, at least.

> The cache check is a quickly added convenience to resolve the current situation of the missing dlls

I think the cache check is actually ingenious and should stay! If the posted versions that AMD uses becomes more stable for future use, comparing the md5 of the cache to the original before blindly accepting its existence could be a good sanity check to avoid a poisoned cache. If the md5 checks out, there's no need to make network requests every time the game is launched

loathingKernel:
> I think the cache check is actually ingenious and should stay!

The cache is checked and used later normally, right before deciding whether to download or not. This is here solely as a means to handle the current situation for users that already had cached dlls.

> If the md5 checks out, there's no need to make network requests

There are already network requests happening before the game is launched, like checking if the manifest for the other upscalers has been updated, which is actually downloaded every time if there is a network connection. FWIW, the check here is not downloading any data, it just checks if the file is available so it's not something that costs an important amount of time to have, and informs us on how to proceed under normal circumstances. The [timeout here](https://github.com/Open-Wine-Components/umu-protonfixes/blob/master/upscalers.py#L72) could be reduced to 1 second probably without it becoming an issue.

loathingKernel (approved):

loathingKernel (changes requested):
I approved this by mistake, sorry. My original reasoning can be found there https://github.com/Open-Wine-Components/umu-protonfixes/pull/526#issuecomment-4116416005

PR #528, opened 2026-03-26
make: build procps-ng without ncurses, fixes failure in steamrt4

loathingKernel:


GloriousEggroll:
ah sorry i actually did this locally and meant to push it

PR #537, opened 2026-04-03
add fix for Gray Zone Warfare EAC hash errors after subseqent launches

ProjectSynchro:
For some reason when you launch the game under Proton these two cache files are either changed or regenerated in some way which causes EAC to boot you from the game after subsequent launches. 

`0xaf497c273f87b6e4_0x7a22fc105639587d.dat`
`0xb9af63cee2e43b6c_0x3cb3b3354fb31606.dat`

The best way to get around this is to set those two files as read only so the game won't invalidate it's own EAC hashes.

I have manually applied this fix for the last few patches and the filenames haven't changed.

Obviously this is a symptom of some syscall not behaving as expected or being implemented incorrectly on the Wine side of things but this should resolve the issue until that is either looked into or resolved.

### DIFF
--- a/gamefixes-steam/2479810.py
+++ b/gamefixes-steam/2479810.py
@@ -1,0 +1,34 @@
+"""Game fix for Gray Zone Warfare
+
+Sets specific cache files to read-only to prevent the game from modifying them
+after first launch, which would invalidate Easy Anti-Cheat hashes.
+"""
+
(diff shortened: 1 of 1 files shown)


ProjectSynchro:
Looks like CI failed due to another issue unrelated to this change.

PR #557, opened 2026-05-23
protonfixes: do not export `PROTON_DLL_COPY=*` on module import

loathingKernel:
`protonfixes` exports `PROTON_DLL_COPY=*` on module import, which causes proton to copy all files from proton's default prefix instead of symlinking them. This causes the prefix to explode in size, delays startup and unnecessarily amplifies disk writes. The reason this is done though, as far back as I could find, was to allow various redist installers to work, including but likely not limited to `dotnet`.

This change removes the export from the module import scope, and instead uses an early-stage fix directive to export the environment variable only for fixes that install a version of dotnet. The scope of the files copied by `PROTON_DLL_COPY` can be reduced further at a later date.

Presently I am not aware of any other redists that require this fix, but we can expand on it later on.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -9,6 +9,7 @@ from . import fix
 from .logger import log
 from .upscalers import setup_upscalers
 from .utilities import (
+    check_verb_requirements,
     setup_frame_rate,
     setup_local_shader_cache,
     setup_mount_drives,
@@ -65,6 +66,7 @@ def execute_early() -> None:
     elif not check_conditions():
         log.warn('Skipping fix execution. We are probably running a unit test.')
     else:
+        check_verb_requirements()
         fix.early()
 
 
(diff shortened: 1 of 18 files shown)


BreiteSeite:
> The reason this is done though, as far back as I could find, was to allow various redist installers to work, including but likely not limited to dotnet.

Any background on why some/"the dotnet" installer won't work with symlinks? Symlinks should be transparent to the applications?

loathingKernel:
> Any background on why some/"the dotnet" installer won't work with symlinks? Symlinks should be transparent to the applications?

I have no idea, and I have not looked into it. This information comes from a private discord server we share with the projects that use umu. I don't think anyone ever looked into it, and at this point I do not care about it enough to do it. If it's needed for a few things, so be it.

BananaWorks07:
why not add this to protontricks instead?

PR #571, opened 2026-06-14
Add game fixes for several id Tech games.

SinnohConfirmed:
Most id Tech 1-3 games will fail to launch on modern GPUs without an environment variable. A few other games that originally used Redbook CD audio need a dll override to use a wrapper that enables in-game music. This wrapper is exclusive to the GOG versions.

I've added fixes for the following games:
Medal of Honor: Allied Assault War Chest
Soldier of Fortune: Platinum Edition
Quake
Hexen II
Star Wars Jedi Knight: Jedi Academy
Star Wars Jedi Knight II: Jedi Outcast
Return to Castle Wolfenstein

Jedi Outcast and Jedi Academy were already on the umu-database, all other games depend on https://github.com/Open-Wine-Components/umu-database/pull/145

I tested all of the id Tech games I own on GOG. I found Quake III and Wolfenstein Enemy Territory do not need any fixes. There are a few more id Tech games on GOG that may need protonfixes that I don't own.

### DIFF
--- a/gamefixes-gog/umu-1207659126.py
+++ b/gamefixes-gog/umu-1207659126.py
@@ -1,0 +1,9 @@
+"""Game fix for Medal of Honor: Allied Assault War Chest"""
+
+from protonfixes import util
+
+
+def main() -> None:
+    # Fix game not launching
+    util.set_environment('MESA_EXTENSION_MAX_YEAR', '2002')
+    util.set_environment('__GL_ExtensionStringVersion', '17700')
--- a/gamefixes-gog/umu-1828104558.py
+++ b/gamefixes-gog/umu-1828104558.py
@@ -3,6 +3,6 @@
 from protonfixes import util
 
 
-def early() ->  None:
+def early() -> None:
     # Fix game not launching
     util.set_environment('PROTON_OLD_GL_STRING', '1')
--- a/gamefixes-gog/umu-1828104558.py
+++ b/gamefixes-gog/umu-1828104558.py
@@ -1,0 +1,9 @@
+"""Game fix for Soldier of Fortune: Platinum Edition"""
+
+from protonfixes import util
+
+
+def main() -> None:
+    # Fix game not launching
+    util.set_environment('MESA_EXTENSION_MAX_YEAR', '2003')
+    util.set_environment('__GL_ExtensionStringVersion', '17700')
--- a/gamefixes-gog/umu-1828104558.py
+++ b/gamefixes-gog/umu-1828104558.py
@@ -3,7 +3,6 @@
 from protonfixes import util
 
 
-def main() -> None:
+def early() ->  None:
     # Fix game not launching
-    util.set_environment('MESA_EXTENSION_MAX_YEAR', '2003')
-    util.set_environment('__GL_ExtensionStringVersion', '17700')
+    util.set_environment('PROTON_OLD_GL_STRING', '1')
--- a/gamefixes-gog/umu-2310.py
+++ b/gamefixes-gog/umu-2310.py
@@ -3,10 +3,11 @@
 from protonfixes import util
 
 
-def main() -> None:
+def early() ->  None:
     # Fix game not launching when using GLQuake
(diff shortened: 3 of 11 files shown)


R1kaB3rN:
Hey, thanks for this.

I'm unable to confirm this on my end, but I looked into this a bit. According to [PCGamingWiki article on Quake](https://www.pcgamingwiki.com/wiki/Quake), it's recommended to delete the `opengl32.dll` file to prevent issues on modern systems. For the [Steam version, that file does not seem to exist](https://steamdb.info/depot/2311/) for it. Have you already confirmed if deleting that has any effect?

SinnohConfirmed:
It seems deleting `opengl32.dll` doesn't have an effect, at least in Proton. With `opengl32.dll` deleted Quake still doesn't launch without `MESA_EXTENSION_MAX_YEAR=2003` and `__GL_ExtensionStringVersion=17700` env variables. By default it doesn't seem like it's attempting to launch with the 3dfx MiniGL wrapper at all under Proton. In SteamDB in the [Steam Deck Compatibility section](https://steamdb.info/app/2310/info/) it shows that the game has the `oldglstr` compat config string, so the Steam version has the same fix being applied by default.

Actually almost all id Tech games with this opengl extension limit issue have the `oldglstr` string in SteamDB. Jedi Outcast, Jedi Academy and Return to Castle Wolfenstein should be pretty much identical on GOG and Steam so I placed their fixes in the `gamefixes-steam` folder and linked them to the `gamefixes-gog` folder. This is kinda redundant since the Steam versions have this fix applied by default. I wasn't sure what the preferred way of adding these fixes were.

R1kaB3rN:
Does setting `PROTON_OLD_GL_STRING`, which is specified in [Proton’s README](https://github.com/ValveSoftware/Proton/blob/proton-11.0-1-beta5/README.md), achieve the same effect as setting the Mesa environment variables? If so, I would prefer setting that instead.

SinnohConfirmed:
`PROTON_OLD_GL_STRING` has the same effect as `MESA_EXTENSION_MAX_YEAR=2003`, but for some reason it doesn't work when in the umu script. Adding `util.set_environment('PROTON_OLD_GL_STRING', '1')` and commenting out the other 2 env variables cause the games to fail to launch, but if I add the env variable outside of Umu then the games work. A real head-scratcher. I should also point out that Medal of Honor needs a value of `2002` to work, so `PROTON_OLD_GL_STRING` doesn't work with that specific game.

imLinguin:
Its because proton parses main after that evaluation. You could use early() to set it up

SinnohConfirmed:
Yep, using early() works! Thanks for the tip :)

I updated my fixes including my Quake II fix that was merged a few days ago.

R1kaB3rN (approved):

gamefixes-gog/umu-1207659126.py
```diff
@@ -0,0 +1,9 @@
+"""Game fix for Medal of Honor: Allied Assault War Chest"""
+
+from protonfixes import util
+
+
+def main() -> None:
+    # Fix game not launching
+    util.set_environment('MESA_EXTENSION_MAX_YEAR', '2002')
```
R1kaB3rN, line 8: Non-blocking: Since the value `2003` did not work from your testing, for future fixes and documentation purposes, I think it would be helpful to know how you determined `2002`.

R1kaB3rN:
Merging, thanks for this.

I would also be open to PRs for other titles that have `oldglstr` set in `proton_compat_config`.

gamefixes-gog/umu-1207659126.py
SinnohConfirmed: I found that `2002` works from [this guide on Steam](https://steamcommunity.com/app/1675200/discussions/0/5267542371396281732/). It's also documented on [this report on Wine's appdb](https://appdb.winehq.org/objectManager.php?sClass=version&iId=20507). [This separate report](https://appdb.winehq.org/objectManager.php?sClass=version&iId=20075) documents that `__GL_ExtensionStringVersion=17700` works on Nvidia drivers.

PR #589, opened 2026-06-24
gamefixes-{egs,umu}: Add fix for Elder Scrolls Online launcher

CommandMC:
Log of the launcher failing:  [eso_launcher.log](https://github.com/user-attachments/files/29306034/eso_launcher.log)
Edit: Updated log with GE-Proton11-1: [launch.log](https://github.com/user-attachments/files/29397102/launch.log)

The game was apparently already added to the DB, so no PR needed there

### DIFF
--- a/gamefixes-umu/umu-306130.py
+++ b/gamefixes-umu/umu-306130.py
@@ -1,0 +1,1 @@
+../gamefixes-egs/umu-306130.py
--- a/gamefixes-egs/umu-306130.py
+++ b/gamefixes-egs/umu-306130.py
@@ -1,0 +1,8 @@
+"""Fix for the Elder Scrolls Online"""
+
+from protonfixes import util
+
+
+def main() -> None:
+    # The launcher is a webview, it does not open correctly without webview2 installed
+    util.protontricks('webview2')


GloriousEggroll:
afaik this should be added as a steam fix (as it exists on steam too with the launcher) then symlinked in the gog folder. there is also a standalone version that should be symlinked in gamefixes-umu and added to the database

CommandMC:
Apologies for the force pushes; just got up & clearly not all there yet
Everything should now be fine

R1kaB3rN:
According to the log file, GE-Proton10-32 was used to launch this game. In future, please submit logs using the latest build of GE-Proton. Also, I looked into this a bit, and, according to various user reports, the game's launcher is known to be problematic. However, for the Steam version, users haven't reported the launcher not opening correctly. In one report in upstream's issue tracker, which was several months ago, the launcher is observed to get put in the foreground. Moreover, this is verified as Playable by the Steam Compatibility Program, and if we were to trust a recent ProtonDB report too, it apparently works out of the box using Proton 10.0-3. Therefore, I would prefer to have this fix scoped to the EGS version until there's confirmation that installing webview2 fixes the Steam version too.

- https://www.protondb.com/app/306130
- https://github.com/ValveSoftware/Proton/issues/556

CommandMC:
Removed the Steam fix & added the standalone version (needs <https://github.com/Open-Wine-Components/umu-database/pull/148>)

Also uploaded a log file with GE-Proton11-1

gamefixes-umu/umu-306130.py
```diff
@@ -0,0 +1 @@
+../gamefixes-egs/umu-306130.py
```
R1kaB3rN, line 1: To confirm, does the standalone version not require any dependencies installed? If dependencies are required (e.g., vcrun2022), then webview2 alone will not suffice and the remaining dependencies need to be installed. According to steamdb.info, VC 2010 Redist is declared as a dependency for this title. Also, currently for all egs titles, protonfixes will install vcrun2022.

- https://github.com/Open-Wine-Components/umu-protonfixes/commit/9361196ec5466a2df3e4e54fedd0c5fcf08f87b7
- https://steamdb.info/app/306130/depots/

R1kaB3rN (changes requested):

gamefixes-umu/umu-306130.py
CommandMC: Sorry, forgot about testing this.
I downloaded the launcher from <https://elderscrolls-a.akamaihd.net/products/BNA_Launcher/Install_ESO.exe>, created a clean prefix, and ran the launcher. It failed with an error message (`Host_IEVersionFail`).

I then installed `webview2`, which made the installer run. In the installer, I have the option of installing "DirectX" and "VC++ 2022 Redistributable" (strangely not ticked by default, though).

Ticking those and clicking "Install", the installer continues and the launcher eventually opens up. I have not tested whether the actual game works (as I'd like to avoid downloading >100GiB for this), but I assume it would.

Do you think we should have DirectX and VCRun as dependencies in this case, even though the installer offers to install them?

R1kaB3rN: Great question, and thanks making the time to follow up.

In this case,  the appropriate action is to not install the dependencies as the installer provides the option of installing them for the game to launch. 

This is a project that primarily expects launcher executables, not installers, and applying fixes with the expectation that the launcher executable (e.g., `zosSteamStarter.exe`) will be run via `wine`. If the installer did not install the dependencies automatically during the setup stage or did not optionally provide to install them in the UI and the game had failed to launch, then I would be in favor of installing them.

R1kaB3rN (approved):

PR #590, opened 2026-06-25
Remove outdated vars. for Wuthering Waves

NelloKudo:
`SteamOS=1` / `SteamDeck=1` no longer allow the game to work since the 3.4 update, but all downstream Proton forks now carry patches to run it anyway. On top of that, the SteamOS/SteamDeck env vars force the game to load a broken Unreal Engine video plugin, resulting in no video during certain cutscenes.

Also removed KRSDK-related override, which is no longer in the game.

### DIFF
--- a/gamefixes-steam/3513350.py
+++ b/gamefixes-steam/3513350.py
@@ -6,10 +6,12 @@ from protonfixes import util
 
 
 def main() -> None:
-    """In-game browser fix."""
-    util.wineexe_override('KRSDKExternal', util.OverrideOrder.DISABLED)
     """Font fixes for in-game resources, if any."""
     util.protontricks('sourcehansans')
     util.protontricks('fakechinese')
     util.protontricks('corefonts')
-    util.set_environment('SteamOS', '1')
+
+    """As of version 3.4, SteamOS/SteamDeck env vars force a broken plugin in-game to play videos."""
+    """Unsetting them to also prevent issues on Steam Decks where they are set by default."""
+    util.del_environment('SteamOS')
+    util.del_environment('SteamDeck')


GloriousEggroll:
? I just launched the game yesterday, is it an updated that has been added since then?

NelloKudo:
> ? I just launched the game yesterday, is it an updated that has been added since then?

Nope, it's been there since the latest 3.4 patch, you probably just didn't notice the issues as they're not immediately obvious. You can easily verify yourself that SteamOS / SteamDeck env vars don't affect the game loading or not in any way anymore and it will keep working on GE-Proton.

As for videos, since this doesn't affect *all* cutscenes but only a few, here's a SteamOS vs non-SteamOS comparison:

- SteamOS=1

https://github.com/user-attachments/assets/b7a1d839-b145-4dfc-84aa-71e501fd5039

- No SteamOS / SteamDeck env. vars set

https://github.com/user-attachments/assets/fd3a7a8d-c19e-4f30-a53c-ba87247e079d

---

Look from 00:13 to 00:16 in the two videos, you'll see in the first one the cutscene stays black while it plays just fine in the second. That is not due to a Wine bug, but it's the game forcing the Unreal Engine `FFMPEGMedia` plugin when such env. vars are set, and that plugin has issues itself for which some videos don't play and others don't loop (also on Windows).

Removing the env vars and keeping the Windows-like behavior instead restores working playback in every case.

gamefixes-steam/3513350.py
```diff
@@ -6,10 +6,12 @@
...
 def main() -> None:
-    """In-game browser fix."""
-    util.wineexe_override('KRSDKExternal', util.OverrideOrder.DISABLED)
     """Font fixes for in-game resources, if any."""
     util.protontricks('sourcehansans')
     util.protontricks('fakechinese')
     util.protontricks('corefonts')
-    util.set_environment('SteamOS', '1')
+
+    """As of version 3.4, SteamOS/SteamDeck env vars force a broken plugin in-game to play videos."""
```
R1kaB3rN, line 14: If this is true, besides unsettling the `SteamOS` and `SteamDeck`, do you know if there is another way to prevent the broken plugin from being loaded?

Users are reportedly setting `SteamOS` and `SteamDeck` in different letter casings with success, so I think preventing the loading of the plugin without unsetting those environment variables would be better.

gamefixes-steam/3513350.py
```diff
@@ -6,10 +6,12 @@
...
+    """Unsetting them to also prevent issues on Steam Decks where they are set by default."""
+    util.del_environment('SteamOS')
```
R1kaB3rN, line 16: To clarify, setting either `SteamOS=1` and `SteamDeck=1` causes the broken plugin to be loaded?

In your first video, you claimed to set only `SteamOS=1` and unset `SteamDeck=1`, then in the second video you unset both `SteamOS=1` and `SteamDeck=1`?

NelloKudo: Yes, either `SteamOS=1` or `SteamDeck=1` alone is enough to trigger the breakage, independently of each other. In the first video I only mentioned SteamOS=1 because that's the one protonfixes set by default.

gamefixes-steam/3513350.py
NelloKudo: I have indeed been looking for a better way but sadly this seems to belong to game code itself.
Unreal Engine games can rely on a setting called `DesiredPlayerName` to override the preferred plugin to use for media and that's what the game seems to be doing. Disabling the related dlls ends up in the game not even loading.

However thanks for the letter casing insight, as it indeed seems that the check is case-insensitive and even cases like `StEaMdEcK=1` still cause the issue... Guess we'll need a better check for this.

R1kaB3rN (approved):

R1kaB3rN:
Merging. This change suffices; I wouldn't concern myself about performing a case insensitive search for the Steam env. variables. The developers need to fix the plugin, and, now, users need to not follow advice that instructs setting those env. variables.

PR #602, opened 2026-07-05
add mfc42 fix for wizardry 8 steam/gog

bleggett:
Wizardry 8 needs mfc42 for the settings dialog to launch.

Tested/confirmed with proton11-cachyos.

xref: https://github.com/Open-Wine-Components/umu-database/pull/149

### DIFF
--- a/gamefixes-gog/245450.py
+++ b/gamefixes-gog/245450.py
@@ -1,0 +1,1 @@
+../gamefixes-steam/245450.py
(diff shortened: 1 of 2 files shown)


loathingKernel:
The fix for GOG in this PR has a malformed filename for the symlink

CI complains about the lack of the `umu-` prefix in the symlink, it should be `gamefixes-gog/umu-245450.py`

PR #613, opened 2026-07-26
Disable NTSync for AC Black Flag Resynced

UsernamesAreNotMyThing:
Apparently Proton 11 has a regression causing the game to not start unless NTSync is disabled.

ValveSoftware/Proton#9958

### DIFF
--- a/gamefixes-steam/3751950.py
+++ b/gamefixes-steam/3751950.py
@@ -1,0 +1,8 @@
+"""Assassin's Creed: Black Flag Resynced"""
+
+from protonfixes import util
+
+
+def main() -> None:
+    """Works around a regression in Proton 11, which is causing the game to crash unless NTSync is disabled."""
+    util.disable_ntsync()


Kishi85:
Note: This NTsync issue seems to be just with starting the Ubisoft Connect Launcher and therefore also affects any other games that use/require it to start the main executable (like e.g. AC Odyssey, Immortals Fenix Rising, etc.)

UsernamesAreNotMyThing:
Since it affects a launcher used by a large number of games, would it just be easier to check if the launcher's install folder is present in the prefix and put it in the initalization logic, instead of putting it in individual gamefixes? That would also make it easier to remove in the case the regression gets fixed.

GloriousEggroll:
[commit 80261c9d](https://github.com/GloriousEggroll/proton-ge-custom/commit/80261c9d617391a33eec877b93353db54b933d9e), included in GE-Proton11-6, addresses Ubisoft Connect’s excessive startup delay caused by redundant split locks, as far as my testing goes it works with ntsync after the ubisoft connect commit fix.

PR #614, opened 2026-07-28
add fix for dungeon town

Renari:


Renari:
It's worth mentioning that I believe this issue is present on Windows as well, meaning their bundled fonts are also not used there.

PR #627, opened 2026-08-16
Add fix for Tom Clancy's Rainbow Six

SinnohConfirmed:
Adds a fix that allows the game to launch in Proton and a fix to allow the game's music to play. To fix the game not launching the game requires the Windows version to be set to Windows 95 or 98 in <code>winecfg</code>. Winetricks currently doesn't support setting these Windows versions in a 64-bit prefix. By editing the prefix's registry directly it's possible to set the Windows version to win98 and this allows the game to launch. This does feel like a hacky solution and I'm open to any solution that is more elegant.


SinnohConfirmed:
depends on https://github.com/Open-Wine-Components/umu-database/pull/152

PR #634, opened 2026-09-06
The Sims 4 (1222670): install the EA app so link2ea:// has a handler

reeveng:
This fixes The Sims 4 not launching at all. The game gets started as a `link2ea://` URI and nothing handles that scheme until the EA app is installed in the prefix, so wine just exits after about ten seconds with no window and nothing useful in the logs.

Turns out steam does try to install it, the game's install script runs the bundled `EAappInstaller.exe`, but without a display mode, so it opens a window and waits for a click that never comes in game mode. Running the same installer with `/quiet` installs it fine and then everything works.

Tested on a Legion Go with Proton 11.0-2. I haven't tried it under umu or GE since that machine is on stock Proton, so let me know if anything needs changing.

More detail and a proton log: https://github.com/ValveSoftware/Proton/issues/4174


imLinguin:
I dont think this varrants a protonfix, Steam should install EA App properly, if it doesnt show you the window - thats a separate problem with SteamOS that has to be sovled.

